PR #1, opened 2018-06-20
Benchmarks test

imakhlin:
- Populate TSDB instance with random data in 24h interval
- Prerequisites: 
1. `v3io.yaml` configuration
2.  existing TSDB instance


yaronha:
@imakhlin we need a seperate dir for that function so it wont conflict with the other function

imakhlin:
Closing and replacing with another PR

PR #2, opened 2018-07-01
Benchmark Test

imakhlin:
- Populate TSDB instance with random data in 24h interval
- Prerequisites: 
1. properly defined `v3io.yaml` configuration
2. existing TSDB instance

> Note: this test provides dynamic capabilities for creating test data in given TSDB instance


test/benchmark/README.md
```diff
@@ -0,0 +1,155 @@
...
+    if [ "$1" == "" ]; then
+      TSDB_PATH="pmetric"
+    else
+      TSDB_PATH=$1
+    fi
+    
+    echo "Creating TSDB instance at $TSDB_PATH"
+    
+    # Create TSDB instance "tsdb" with some pre-aggregations and interval of 5 seconds (-v for verbose mode)
+    SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
```
gtopper: Do we really have to `cd` in there?

test/benchmark/BenchmarkIngestWithNuclio_test.go
```diff
@@ -0,0 +1,148 @@
+package benchmark
```
gtopper: I don't think it's proper for benchmark to be used both as a source root and a package.

test/benchmark/BenchmarkIngestWithNuclio_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	testConfig, err := common.LoadBenchmarkIngestConfigFromData()
+	if err != nil {
+		panic(err)
+	}
+
+	v3ioConfigFile := os.Getenv("V3IO_TSDBCFG_PATH")
+	v3ioConfig, err := config.LoadConfig(v3ioConfigFile)
+	if err != nil {
+		panic(errors.Wrap(err, fmt.Sprintf("Failed to load config from file %s", v3ioConfigFile)))
```
gtopper: > Error strings should not start with a capital letter because they'll often be prefixed before printing:
https://github.com/golang/go/wiki/Errors
(yes, I know this is violated in a bunch of other places in the source code)

test/benchmark/common/config.go
```diff
@@ -0,0 +1,56 @@
...
+func LoadBenchmarkIngestConfigFromData() (*BenchmarkIngestConfig, error) {
+	var benchConfigFile = os.Getenv(TsdbBenchIngestConfig)
+	if benchConfigFile == "" {
+		benchConfigFile = "tsdb-bench-test-config.yaml"
+	}
+
+	configData, err := ioutil.ReadFile(benchConfigFile)
+	if err != nil {
+		log.Fatal(err)
+		os.Exit(2)
```
gtopper: I think this is when you're supposed to panic 😱

gtopper: Or, alternatively, fail the test.

gtopper: I don't think you're supposed to system exit on the test suite 😅

test/benchmark/BenchmarkIngest_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	// run the runTest function b.N times
+	relativeTimeOffsetMs, err := utils.Str2duration(testConfig.StartTimeOffset)
+	if err != nil {
+		b.Fatal("Unable to resolve start time. Check configuration.")
+	}
+	testStartTimeMs := testStartTimeNano/int64(time.Millisecond) - relativeTimeOffsetMs
+	timeStampsCount := (testStartTimeNano/int64(time.Millisecond) - testStartTimeMs) / int64(testConfig.SampleStepSize)
+	timeStamps := make([]int64, timeStampsCount)
```
gtopper: `timeStamps` -> `timestamps`

test/benchmark/common/model.go
```diff
@@ -0,0 +1,131 @@
...
+	for name, labels := range model {
+		valSetLenght := 0
+		// TODO: find better way to get size of the label values array
+		for _, labelValues := range labels {
+			valSetLenght = len(labelValues)
+			break
+		}
+		for index := 0; index < valSetLenght; index++ {
+			var buffer bytes.Buffer
+			buffer.WriteString(fmt.Sprintf("{\"Lset\": { \"__name__\":\"%s\"", name))
```
gtopper: You can use ticks instead: `` `{"xyz":123}` ``

test/benchmark/common/model.go
```diff
@@ -0,0 +1,131 @@
...
+			}
+		}
+	}
+	return model
+}
+
+func MakeSampleTemplates(model map[string]map[string][]string) []string {
+	var result = make([]string, 0)
+	for name, labels := range model {
+		valSetLenght := 0
```
gtopper: Typo

test/benchmark/common/model.go
```diff
@@ -0,0 +1,131 @@
...
+		// TODO: find better way to get size of the label values array
+		for _, labelValues := range labels {
+			valSetLenght = len(labelValues)
```
gtopper: Shouldn't you be adding rather than replacing?

test/benchmark/common/model.go
```diff
@@ -0,0 +1,131 @@
...
+			for _, labelValues := range labelValues {
+				model[name][label] = append(model[name][label], labelValues)
+			}
+		}
+	}
+	return model
+}
+
+func MakeSampleTemplates(model map[string]map[string][]string) []string {
+	var result = make([]string, 0)
```
gtopper: Zero is a pretty lousy guess 🙊0️⃣

test/benchmark/common/model.go
gtopper: Better yet, you could marshal this stuff rather than unsafely concat strings.

test/benchmark/common/model.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	limit := maxIndex - minIndex + 1
+	size := 1
+	if count == 1 {
+		size = limit
+	} else {
+		size = count * limit
+	}
+	array := make([]string, size)
```
gtopper: It's not an array, it's a slice 😄

test/benchmark/BenchmarkIngest_test.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func runTest(
+	index int,
+	appender tsdb.Appender,
+	testCtx *testing.B,
+	timeStamps []int64,
+	sampleTemplates []string,
+	flushFrequency int,
+	refsMap map[uint64]bool,
```
gtopper: Doesn't look like this needs to be a map. We always save `true` in the value and never use it...

test/benchmark/BenchmarkIngest_test.go
```diff
@@ -0,0 +1,192 @@
...
+	sample := common.JsonTemplate2Sample(sampleTemplateJson, testCtx, timeStamp, common.MakeRandomFloat64())
+
+	if cycleId == 0 {
+		// initialize refIds
+		// Add first & get reference
+		ref, err := appender.Add(sample.Lset, timeStamp, sample.Value)
+		if err != nil {
+			testCtx.Fatalf("Add request has failed!\nError: %s", err)
+		}
+		refsArray[refIndex] = ref
```
gtopper: This too is not an array 😄

test/benchmark/BenchmarkIngest_test.go
```diff
@@ -0,0 +1,192 @@
...
+		ref, err := appender.Add(sample.Lset, initialTimeStamp, sample.Value)
+		if err != nil {
+			testCtx.Fatalf("Add request has failed!\nError: %s", err)
+		}
+		refsArray[i] = ref
+		refsMap[ref] = true
+		count++
+	}
+
+	timeSerieSize := len(timeStamps)
```
gtopper: `timeSerieSize` -> `timeSeriesSize`

test/benchmark/BenchmarkIngest_test.go
```diff
@@ -0,0 +1,192 @@
...
+	for dataPointIndex := 1; dataPointIndex < timeSerieSize; dataPointIndex++ {
+		for refIndex, sampleTemplateJson := range sampleTemplates {
+			sample := common.JsonTemplate2Sample(sampleTemplateJson, testCtx, timeStamps[dataPointIndex], common.MakeRandomFloat64())
+			err := appender.AddFast(sample.Lset, refsArray[refIndex], timeStamps[dataPointIndex], sample.Value)
+			if err != nil {
+				testCtx.Fatalf("AddFast request has failed!\nSample:%v\nError: %s", sample, err)
```
gtopper: I'm pretty sure you can use `errors.Wrap` here instead of these fussy strings.

test/benchmark/README.md
imakhlin: yes, because it need the full path

test/benchmark/common/model.go
imakhlin: doing it only once

test/benchmark/common/model.go
imakhlin: I have no better number at this point ;)

test/benchmark/README.md
gtopper: Use `realpath`?

PR #3, opened 2018-07-04
Change array separator to match documentation (`;` -> `,`).

gtopper:
Also:
* Minor improvement to usage string.
* Use MixedCaps as per Go convention.

### DIFF
--- a/pkg/tsdbctl/add.go
+++ b/pkg/tsdbctl/add.go
@@ -36,7 +36,7 @@ import (
 	"github.com/v3io/v3io-tsdb/pkg/tsdb"
 )
 
-const ARRAY_SEPARATOR = ";"
+const ArraySeparator = ","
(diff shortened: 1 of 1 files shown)


yaronha:
@gtopper we need to change the other way around, that was modified to `;` so it will make sense in CSV w/o extra quotes

gtopper:
Sure, but shouldn't this be consistent also for labels? They are currently comma-separated.

gtopper:
Also, semicolons are not so shell friendly :-)

gtopper:
@yaronha: Effective Go says [MixedCaps](https://golang.org/doc/effective_go.html#mixed-caps). Our Nuclio sources seem to mix MixedCaps with camelCase. SNAKE_CASE is probably not really a Go thing.

PR #8, opened 2018-07-08
Ran `go fmt ./...`

gtopper:


imakhlin:
Please make sure that `go fmt` is aligned with IDE (GoLand)

gtopper:
Turns out that GoLand has different alignment rules (maybe in line with an older version of Go?). Anyway, I think we better stick with the toolchain.

imakhlin:
IMPORTANT NOTE: After merging #8 and #9 contributors will be required to run `go fmt ./...` before pushing to the repo and not rely on the IDE formatting since there are some differences

PR #15, opened 2018-07-11
Minor restructuring toward a more standard project.

gtopper:
With related and incidental documentation updates.

Did not move `promtsdb` to avoid breaking [github.com/v3io/prometheus](github.com/v3io/prometheus).


imakhlin:
LGTM

PR #16, opened 2018-07-11
Exclude vendor/ in Makefile targets.

gtopper:


Makefile
```diff
@@ -1,11 +1,16 @@
 .PHONY: test
+
+# All top-level dirs except for vendor/.
+TOPLEVEL_DIRS=`ls -d ./*/. | grep -v '^./vendor/.$$' | sed 's/\.$$/.../'`
```
imakhlin: Typo, should be `TOPPER_LEVEL`...
:)

PR #19, opened 2018-07-19
Master to dev

gtopper:


pkg/tsdbctl/add.go
```diff
@@ -55,7 +55,7 @@ func newAddCommandeer(rootCommandeer *RootCommandeer) *addCommandeer {
 	}
 
 	cmd := &cobra.Command{
-		Use:     "add metric [labels] [flags]",
+		Use:     "add <metric> [labels] [flags]",
```
imakhlin: We need to use same syntax everywhere, I mean <xxx>

PR #28, opened 2018-07-30
Add tests

talIguaz:
Added also three build tags:
`unit` - for all unit tests (run `make test`)
`integration` - for all integration tests (run `make integration`)


pkg/chunkenc/chunkenc_test.go
```diff
@@ -100,6 +103,8 @@ func TestXor(tst *testing.T) {
 }
 
 func TestBstream(t *testing.T) {
+	t.Skip("Needs to be refactored - Doesnt test anything")
```
gtopper: Doesn't :)

pkg/aggregate/aggregate_test.go
```diff
@@ -1,20 +1,81 @@
...
+			fmt.Sprintf("_v_sum[1]=%f;", 10.0), false},
+
+		{"Should aggregate data with Sqr aggregator", "sqr", map[int64]float64{1: 2.0}, "v", 1,
+			fmt.Sprintf("_v_sqr[1]=_v_sqr[1]+%f;", 4.0),
+			fmt.Sprintf("_v_sqr[1]=%f;", 4.0), false},
+
+		// todo: enable when bug is fixed - IG-8675
+		//{"Should aggregate data with Min & Max aggregators", "min,max", map[int64]float64{1: 7.5, 2: 2.5}, "v", 1,
+		//	fmt.Sprintf("_v_min[1]=min(_v_min[1],%f);_v_max[1]=max(_v_max[1],%f);", 2.5, 7.5),
+		//	fmt.Sprintf("_v_min[1]=%f;_v_max[1]=%f;", 2.5, 7.5), false},
```
gtopper: This pattern kinda sucks if we have to comment out tests and log them ourselves... I prefer good old code sharing by function (where the tests call the common code and not vice versa).

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+func setUp(t *testing.T, v3ioConfig *config.V3ioConfig) func() {
+	dbConfig := config.DBPartConfig{
+		DaysPerObj:     1,
+		HrInChunk:      1,
+		DefaultRollups: "sum",
+		RollupMin:      10,
+	}
+
+	if err := CreateTSDB(v3ioConfig, &dbConfig); err != nil {
```
gtopper: We need this stuff to be repeatable. When I try to run it, I get
```
--- FAIL: TestQueryData (0.04s)
	v3iotsdb_integration_test.go:220: Should ingest and query one data point
	v3iotsdb_integration_test.go:56: Failed to create TSDB. reason: TSDB already exist in path: pmetric
```

test/utils/types.go
```diff
@@ -0,0 +1,14 @@
+package utils
```
gtopper: `/test` [appears to mean](https://github.com/golang-standards/project-layout#test) external test apps. Seems like stuff that should only be shared between tests (and even that's arguable, as you won't be able to ever modify these without refactoring all the tests - a major pain caused by unnecessary coupling of all the tests).

Makefile
```diff
@@ -19,7 +19,11 @@ get:
 
 .PHONY: test
 test: get
-	go test -v -short $(TOPLEVEL_DIRS)
+	go test -v -short -tags unit $(TOPLEVEL_DIRS)
```
gtopper: Why do we need both `-short` and `-tags`? Would be better to do such things in separate PRs.

examples/nuclio/ingest/ingest_example_test.go
```diff
@@ -1,9 +1,13 @@
+// +build nuclio_integration
```
gtopper: This one doesn't have a make target. Also, not really clear to me what this means... This integrates with a nuclio instance? Where does it find it? Or is this just integration against the sdk? If it's the latter, I don't see why we would want a tag for that. We use all kinds of libraries and don't mention it.

examples/nuclio/ingest/ingest_example_test.go
```diff
@@ -13,9 +17,14 @@ func TestIngestIntegration(t *testing.T) {
 	if testing.Short() {
 		t.Skip("Skipping integration test.")
 	}
+	v3ioConfig, err := config.LoadConfig(filepath.Join("../../../", config.DefaultConfigurationFileName))
```
gtopper: `filepath.Join` takes as many arguments as you wish, use it!

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+	appender, err := adapter.Appender()
+	if err != nil {
+		t.Fatalf("Failed to get appender. reason: %s", err)
+	}
+
+	labels := utils.Labels{utils.Label{Name: "__name__", Value: metricsName}}
+	labels = append(labels, userLabels...)
+
+	fmt.Printf("the labeles are: %v\n", labels)
```
gtopper: Typo. Do we really want to print stuff here? And if so, shouldn't we use the logger?

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -0,0 +1,350 @@
...
+	ref, err := appender.Add(labels, data[0].Time, data[0].Value)
+	if err != nil {
+		t.Fatalf("Failed to add data to appender. reason: %s", err)
+	}
+	for i := 1; i < len(data); i++ {
+		appender.AddFast(labels, ref, data[i].Time, data[i].Value)
```
gtopper: You can range over `data[1:]` instead.

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+	labels := utils.Labels{utils.Label{Name: "__name__", Value: metricsName}}
+	labels = append(labels, userLabels...)
+
+	fmt.Printf("the labeles are: %v\n", labels)
+	ref, err := appender.Add(labels, data[0].Time, data[0].Value)
+	if err != nil {
+		t.Fatalf("Failed to add data to appender. reason: %s", err)
+	}
+	for i := 1; i < len(data); i++ {
```
gtopper: You can range over data[1:] instead.

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -0,0 +1,350 @@
...
+		appender.AddFast(labels, ref, data[i].Time, data[i].Value)
+	}
+
+	if _, err := appender.WaitForCompletion(0); err != nil {
+		t.Fatalf("Failed to wait for appender completion. reason: %s", err)
+	}
+
+	responseChan := make(chan *v3io.Response)
+	adapter.container.GetItems(&v3io.GetItemsInput{
+		Path:           "/metrics/0/",
```
gtopper: Where does this path come from? Your `v3io.yaml`? :)

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -0,0 +1,350 @@
...
+	}, 30, responseChan)
+
+	res := <-responseChan
+	getItemsResp := res.Output.(*v3io.GetItemsOutput)
+
+	for _, item := range getItemsResp.Items {
+		for _, label := range userLabels {
+			actual := item.GetField(label.Name)
+			if actual != label.Value {
+				t.Fatalf("Records were not saved correctly. for label %s, actual: %v, expected: %v",
```
gtopper: `correctly.` -> `correctly`

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	labels := utils.Labels{utils.Label{Name: "__name__", Value: metricsName}}
+	labels = append(labels, userLabels...)
+
+	ref, err := appender.Add(labels, data[0].Time, data[0].Value)
+	if err != nil {
+		test.Fatalf("Failed to add data to appender. reason: %s", err)
+	}
+	for i := 1; i < len(data); i++ {
```
gtopper: range over `data[1:]`?

test/utils/types.go
gtopper: Also not really utils at all... These are common types :)

Makefile
imakhlin: agree with @gtopper. I think we should get rid of the short and use tags for test suites

examples/nuclio/ingest/ingest_example_test.go
imakhlin: store test resources and configuration at `testdata` directory

pkg/tsdb/tsdbtest/tsdbtest.go
```diff
@@ -0,0 +1,57 @@
+package tsdbtest
```
imakhlin: naming convention: `tsdb_test.go`

pkg/tsdb/tsdbtest/tsdbtest.go
```diff
@@ -0,0 +1,57 @@
...
+	Value float64
+}
+
+type Sample struct {
+	Lset  utils.Labels
+	Time  string
+	Value float64
+}
+
+func DeleteTSDB(t *testing.T, v3ioConfig *config.V3ioConfig) {
```
imakhlin: look at /src/github.com/v3io/v3io-tsdb/test/benchmark/common/utils.go
make sure we don't duplicate code

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -0,0 +1,311 @@
...
+	. "github.com/v3io/v3io-tsdb/pkg/tsdb"
+	"github.com/v3io/v3io-tsdb/pkg/tsdb/tsdbtest"
+	"github.com/v3io/v3io-tsdb/pkg/utils"
+	"path/filepath"
+	"reflect"
+	"testing"
+)
+
+func TestIngestData(t *testing.T) {
+	v3ioConfig, err := config.LoadConfig(filepath.Join("../../", config.DefaultConfigurationFileName))
```
imakhlin: Look at the src/github.com/v3io/v3io-tsdb/test/benchmark/common/config.go
Lets unify the code

pkg/aggregate/aggregate_test.go
```diff
@@ -1,20 +1,81 @@
...
+
+		{"Should aggregate data with Sum aggregator", "sum", map[int64]float64{1: 7.5, 2: 2.5}, "v", 1,
+			fmt.Sprintf("_v_sum[1]=_v_sum[1]+%f;", 10.0),
+			fmt.Sprintf("_v_sum[1]=%f;", 10.0), false},
+
+		{"Should aggregate data with Sqr aggregator", "sqr", map[int64]float64{1: 2.0}, "v", 1,
+			fmt.Sprintf("_v_sqr[1]=_v_sqr[1]+%f;", 4.0),
+			fmt.Sprintf("_v_sqr[1]=%f;", 4.0), false},
+
+		// todo: enable when bug is fixed - IG-8675
```
gtopper: Shouldn't be needed anymore with proper ignores, but generally speaking `todo`->`TODO` 👷

examples/nuclio/ingest/ingest_example_test.go
```diff
@@ -1,21 +1,29 @@
...
+	"github.com/v3io/v3io-tsdb/pkg/config"
+	"github.com/v3io/v3io-tsdb/pkg/tsdb/tsdbtest"
 	"os"
+	"path/filepath"
 	"testing"
-	"time"
 )
 
 func TestIngestIntegration(t *testing.T) {
+	v3ioConfig, err := config.LoadConfig(filepath.Join("..", "..", "..", config.DefaultConfigurationFileName))
```
imakhlin: use `testdata` dir to store test configurations

PR #30, opened 2018-07-31
Fix Bench test

imakhlin:
* Fix table path - use underscores 
* Make BatchSize configurable
* Update BenchIngest test

### DIFF
--- a/test/benchmark/BenchmarkIngest_test.go
+++ b/test/benchmark/BenchmarkIngest_test.go
@@ -8,8 +8,6 @@ import (
 	"github.com/v3io/v3io-tsdb/test/benchmark/common"
 	"io/ioutil"
 	"log"
-	"regexp"
-	"strings"
 	"testing"
 	"time"
 )
@@ -28,7 +26,7 @@ func BenchmarkIngest(b *testing.B) {
 	}
 
 	// Create test path (tsdb instance)
-	tsdbPath := formatPath(fmt.Sprintf("tsdb-%s-%d-%s", b.Name(), b.N, time.Now().Format(time.RFC3339)))
+	tsdbPath := common.NormalizePath(fmt.Sprintf("tsdb-%s-%d-%s", b.Name(), b.N, time.Now().Format(time.RFC3339)))
 
 	// Update TSDB instance path for this test
 	v3ioConfig.Path = tsdbPath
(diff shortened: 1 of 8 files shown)


test/benchmark/BenchmarkIngest_test.go
```diff
@@ -112,6 +112,47 @@ func BenchmarkIngest(b *testing.B) {
 	b.Logf("\nTest complete. %d samples added to %s\n", count, tsdbPath)
 
 	// TODO: Make sure all samples have been ingested properly, i.e. execute `count` for all metrics and compare with expected
+	if err := validateCountOfSamples(adapter, count, testStartTimeMs, testEndTimeMs); err != nil {
+		b.Error(err)
+	}
+}
+
+func validateCountOfSamples(adapter *tsdb.V3ioAdapter, expected int, startTimeMs, endTimeMs int64) error {
```
talIguaz: maybe pass interface `TB` (mutual between T and B tests) so that the valudation will fail the test directly instead of checking it on every test

test/benchmark/common/utils.go
```diff
@@ -18,3 +23,49 @@ func CreateTSDB(v3ioConfig *config.V3ioConfig, newTsdbPath string) error {
 func DeleteTSDB(adapter *tsdb.V3ioAdapter, deleteConf bool, force bool) {
 	adapter.DeleteDB(deleteConf, force)
 }
+
+func ValidateCountOfSamples(adapter *tsdb.V3ioAdapter, expected int, startTimeMs, endTimeMs int64) error {
```
talIguaz: Maybe pass interface TB (mutual between T and B tests) so that the valudation will fail the test directly instead of checking it on every test

test/benchmark/BenchmarkIngest_test.go
```diff
@@ -79,20 +93,25 @@ func BenchmarkIngest(b *testing.B) {
 			b.Fatal(err)
 		}
 
-		if rowsAdded == 0 {
+		count += rowsAdded
+
+		if rowsAdded == 0 || rowsAdded >= testLimit {
+			defer b.Skipf("\nTest have reached the target (limit=%d)", testLimit)
```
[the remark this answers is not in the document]
imakhlin: I want this test to finish and then exit the entire bench test (stop, since the limit has reached)

test/benchmark/testdata/tsdb-bench-test-config.yaml
```diff
@@ -34,4 +34,4 @@ AppendOneByOne: true
 BatchSize: 64
 
 # Cleanup the data after each test
-CleanupAfterTest: true
\ No newline at end of file
+CleanupAfterTest: false
```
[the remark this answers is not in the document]
imakhlin: after adding validation - yes

test/benchmark/common/utils.go
imakhlin: in that case you won't be able to use this validation outside of the test (as an API)
if this function is going to be used for tests only we can do it

PR #38, opened 2018-08-06
partition manager fixes

dinal:


pkg/partmgr/partmgr.go
```diff
@@ -97,7 +102,7 @@ func (p *PartitionManager) Init() error {
 }
 
 func (p *PartitionManager) TimeToPart(t int64) (*DBPartition, error) {
-	interval := IntervalInMilli(p.cfg.PartitionSchemaInfo.PartitionerInterval)
+	interval, _ := utils.Str2duration(p.cfg.PartitionSchemaInfo.PartitionerInterval)
```
yaronha: @dinal rather not do string parsing on fast path, need to be taken from the priv variables

pkg/tsdb/tsdbtest/tsdbtest.go
```diff
@@ -63,36 +63,36 @@ func SetUpWithDBConfig(t *testing.T, v3ioConfig *config.V3ioConfig, schema *conf
...
+		LayerRetentionTime:     "1y",
 	}
 
 	tableSchema := config.TableSchema{
 		Version:             0,
 		RollupLayers:        []config.Rollup{defaultRollup},
 		ShardingBuckets:     64,
-		PartitionerInterval: "1D",
-		ChunckerInterval:    "1H",
+		PartitionerInterval: "1d",
```
yaronha: @dinal i would make 2 days the default (so we have 48 chunks)

pkg/tsdb/v3iotsdb.go
```diff
@@ -172,20 +172,21 @@ func (a *V3ioAdapter) Querier(_ context.Context, mint, maxt int64) (*querier.V3i
...
-func (a *V3ioAdapter) DeleteDB(configExists bool, force bool) error {
+func (a *V3ioAdapter) DeleteDB(configExists bool, force bool, fromTime int64, toTime int64) error {
 
-	path := a.partitionMngr.GetHead().GetTablePath()
-	a.logger.Info("Delete partition %s", path)
-	err := utils.DeleteTable(a.container, path, "", a.cfg.QryWorkers)
-	if err != nil && !force {
-		return err
+	partitions := a.partitionMngr.PartsForRange(fromTime, toTime)
+	for _, part := range partitions {
```
yaronha: @dinal i assume we dont need to delete partitions where toTime hits the middle of the partition, so maybe check if the partition.maxTime is < toTime

pkg/tsdbctl/delete.go
```diff
@@ -51,6 +53,8 @@ func newDeleteCommandeer(rootCommandeer *RootCommandeer) *delCommandeer {
 
 	cmd.Flags().BoolVarP(&commandeer.delConfig, "del-config", "d", false, "Delete the TSDB config as well")
 	cmd.Flags().BoolVarP(&commandeer.force, "force", "f", false, "Delete all elements even if some steps fail")
+	cmd.Flags().Int64VarP(&commandeer.fromTime, "from-time", "r", 0, "Delete partitions from this time and on")
+	cmd.Flags().Int64VarP(&commandeer.toTime, "to-time", "t", 0, "Delete partitions up to this time")
 	commandeer.cmd = cmd
```
yaronha: @dinal better use the string representation of time in CLI (can work with multiple date semantics), see query commanf begin/end times

PR #39, opened 2018-08-06
Support NaN in aggregate functions

imakhlin:
1. Init array with NaN
2. Compare Min/Max with NaN

Note: Depends on PR (https://github.com/iguazio/zeek/pull/5757) to the engine (@gals)


gtopper:
We agreed not to use NaN in the end, but rather use each aggregation's logical zero value (±Inf for min/max, zero for the rest). We may also use the count aggregator to determine existence of values.

PR #41, opened 2018-08-07
Partitions 

gshatz:


pkg/aggregate/aggregate.go
```diff
@@ -58,11 +59,51 @@ var aggrToString = map[AggrType]string{
 	aggrTypeStddev: "stddev", aggrTypeStdvar: "stdvar", aggrTypeAll: "*",
 }
 
+var aggrToSchemaField = map[string]config.SchemaField{
+	"count":  {Name: "count", Type: "array", Nullable: true, Items: "double"},
```
imakhlin: I know that the list of supported aggregators is fixed right now but it could change later. I think it will be more flexible if this list will be defined in configuration.

pkg/aggregate/aggregate.go
```diff
@@ -58,11 +59,51 @@ var aggrToString = map[AggrType]string{
...
+	"max":    {Name: "max", Type: "array", Nullable: true, Items: "double"},
+	"min":    {Name: "min", Type: "array", Nullable: true, Items: "double"},
+	"last":   {Name: "last", Type: "array", Nullable: true, Items: "double"},
+	"avg":    {Name: "avg", Type: "array", Nullable: true, Items: "double"},
+	"rate":   {Name: "rate", Type: "array", Nullable: true, Items: "double"},
+	"stddev": {Name: "stddev", Type: "array", Nullable: true, Items: "double"},
+	"stdvar": {Name: "stdvar", Type: "array", Nullable: true, Items: "double"},
+}
+
+func SchemaFieldFromString(split []string, col string) ([]config.SchemaField, error) {
```
imakhlin: "split" -> "aggregators"

pkg/aggregate/aggregate.go
```diff
@@ -58,11 +59,51 @@ var aggrToString = map[AggrType]string{
...
+			if strings.Compare(s, "*") == 0 {
+				fieldList = make([]config.SchemaField, 0, len(aggrToSchemaField))
+				for _, val := range aggrToSchemaField {
+					fieldList = append(fieldList, getAggrFullName(val, col))
+				}
+				return fieldList, nil
+			} else {
+				field, ok := aggrToSchemaField[s]
+				if !ok {
+					return fieldList, fmt.Errorf("Invalid aggragator type %s", s)
```
imakhlin: partial result and error will be returned. Is this what you want?

pkg/aggregate/aggregate.go
```diff
@@ -58,11 +59,51 @@ var aggrToString = map[AggrType]string{
...
+				}
+				fieldList = append(fieldList, getAggrFullName(field, col))
+			}
+		}
+	}
+	return fieldList, nil
+}
+
+func getAggrFullName(field config.SchemaField, col string) config.SchemaField {
+	fullName := fmt.Sprintf("_%s_%s", col, field.Name)
```
imakhlin: for empty `col` it will yield `__FieldName`

pkg/config/config.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 const DefaultConfigurationFileName = "v3io.yaml"
+const SCHEMA_CONFIG = "/.schema"
```
imakhlin: `/` should be removed
Use `filePath.Join(...)` instead (see https://golang.org/src/path/filepath/path.go?s=5895:5927#L199)

pkg/config/config.go
```diff
@@ -54,42 +55,52 @@ type V3ioConfig struct {
...
-	// Metric specific policy
-	MetricsConfig map[string]MetricConfig `json:"metricsConfig,omitempty"`
+type Rollup struct {
+	Aggregators            string `json:"aggregators"`
+	AggregatorsGranularity string `json:"aggregatorsGranularity"`
+	//["cloud","local"] for the aggregators and sample chucks
+	StorageClass string `json:"storageClass"`
+	//in hours. 0  means no need to save samples
+	SampleRetention int `json:"sampleRetention"`
+	// format : 1M, 7D, 3H . Possible intervals: Y/M/D/H/M
```
imakhlin: should use "m" for minutes and "M" for mounts

pkg/config/config.go
```diff
@@ -54,42 +55,52 @@ type V3ioConfig struct {
...
+	LayerRetentionTime string `json:"layerRetentionTime"`
```
imakhlin: Is there any conceptual difference between the `SampleRetention` and `LayerRetentionTime`?
If no then they should use same naming convention, i.e. `LayerRetention`

pkg/operators/mergers.go
```diff
@@ -0,0 +1,110 @@
...
+
+In addition, you may not use the software for any purposes that are
+illegal under applicable law, and the grant of the foregoing license
+under the Apache 2.0 license is conditioned upon your compliance with
+such restriction.
+*/
+
+package operators
+
+type IterSortMerger struct {
```
imakhlin: add unit tests

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
-	mtx           sync.RWMutex
-	path          string
-	cfg           *config.DBPartConfig
-	headPartition *DBPartition
-	cyclic        bool
-	ignoreWrap    bool
+	mtx                      sync.RWMutex
+	path                     string
+	cfg                      *config.Schema
+	headPartition            *DBPartition
```
imakhlin: why not just store the index of head partition?

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
+	container                *v3io.Container
+	currentPartitionInterval int64 //TODO update on schema changes
 }
 
 func (p *PartitionManager) IsCyclic() bool {
 	return p.cyclic
 }
 
-func (p *PartitionManager) GetConfig() *config.DBPartConfig {
+func (p *PartitionManager) GetPartitions() []*DBPartition {
```
imakhlin: You can make it exportable if you want allow others to read and modify the value.

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
+	}
+	return p.headPartition, nil
 }
 
-func (p *PartitionManager) PartsForRange(mint, maxt int64) []*DBPartition {
-
-	return []*DBPartition{p.headPartition}
+func (p *PartitionManager) createNewPartition(t int64) (*DBPartition, error) {
+	time := t & 0x7FFFFFFFFFFFFFF0
+	path := fmt.Sprintf("%s/%d/", p.path, time/1000)
```
imakhlin: use `filepath.Join(...)`

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
-// Create new Partition Manager, for now confined to one Cyclic partition
-func NewPartitionMngr(cfg *config.DBPartConfig, path string) *PartitionManager {
-	newMngr := &PartitionManager{cfg: cfg, path: path, cyclic: true, ignoreWrap: true}
-	newMngr.headPartition = NewDBPartition(newMngr)
+// Create new Partition Manager
+func NewPartitionMngr(cfg *config.Schema, path string, cont *v3io.Container) *PartitionManager {
+	currentPartitionInterval, _ := utils.Str2duration(cfg.PartitionSchemaInfo.PartitionerInterval)
+	newMngr := &PartitionManager{cfg: cfg, path: path, cyclic: false, container: cont, currentPartitionInterval: currentPartitionInterval}
+	for _, part := range cfg.Partitions {
+		path := fmt.Sprintf("%s/%d/", newMngr.path, part.StartTime/1000)
```
imakhlin: use `filepath.Join(...)`

pkg/aggregate/aggregate.go
yaronha: i would skip it now

pkg/aggregate/aggregate.go
yaronha: its never empty

pkg/aggregate/aggregate.go
yaronha: col is never empty (illegal)

pkg/partmgr/partmgr.go
yaronha: i assume indexes may change as partitions can be deleted

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
 }
 
 func (p *DBPartition) IsCyclic() bool {
 	return p.manager.cyclic
 }
 
-func (p *DBPartition) HoursInChunk() int {
-	return p.hoursInChunk
+// Time covered by a single chunk
+func (p *DBPartition) TimePerChunk() int64 {
```
imakhlin: Isn't it `ChunkInterval`? Time is too abstract... If we decided to use same units (Milliseconds) everywhere then lets omit the time/units

pkg/partmgr/partmgr.go
yaronha: why? whats the advantage?

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
+}
+
+// return path to metrics table
+func (p *DBPartition) GetTablePath() string {
+	return p.path
+}
+
+// return list of Sharding Keys matching the name
+func (p *DBPartition) GetShardingKeys(name string) []string {
+	var res []string
```
imakhlin: you can initialise the slice with `len(p.manager.cfg.TableSchemaInfo.ShardingBuckets)`

pkg/querier/multipart.go
```diff
@@ -0,0 +1,217 @@
+package querier
+
+import (
+	"github.com/v3io/v3io-tsdb/pkg/utils"
+	"sort"
+	"strings"
+)
+
+func NewSetSorter(set SeriesSet) (SeriesSet, error) {
```
imakhlin: Add tests

pkg/querier/seriesset.go
```diff
@@ -0,0 +1,244 @@
...
+// Get relevant items & attributes from the DB, and create an iterator
+// TODO: get items per partition + merge, per partition calc attrs
+func (s *V3ioSeriesSet) getItems(partition *partmgr.DBPartition, name, filter string, container *v3io.Container, workers int) error {
+
+	path := partition.GetTablePath()
+	shardingKeys := []string{}
+	if name != "" {
+		shardingKeys = partition.GetShardingKeys(name)
+	}
+	attrs := []string{"_lset", "_ooo", "_name", "_maxtime"}
```
imakhlin: What `_ooo` stands for?

yaronha: @imakhlin out of order

pkg/querier/seriesset.go
```diff
@@ -0,0 +1,244 @@
...
+
+		if iter.Err() != nil {
+			s.err = iter.Err()
+			return
+		}
+
+		t0, _ := iter.At()
+		s.baseTime = (t0 / s.interval) * s.interval
+
+		for {
```
imakhlin: isn't it equivalent to `for ; iter.Next(); {...}`

yaronha: @imakhlin no, we need to init baseTime, but there seem to be a bug (the if iter.Err() need to go out of the if, since .Next will return false in case of error)

pkg/querier/seriesset.go
```diff
@@ -0,0 +1,244 @@
...
+			t, v := iter.At()
+			s.aggrSet.AppendAllCells(int((t-s.baseTime)/s.interval), v)
+			if !iter.Next() {
+				// s.err = iter.error()  // if the internal iterator has error we dont need to err the aggregator
```
imakhlin: please remove the dead code

pkg/tsdb/v3iotsdb.go
```diff
@@ -182,31 +172,34 @@ func (a *V3ioAdapter) Querier(_ context.Context, mint, maxt int64) (*querier.V3i
...
 
-func (a *V3ioAdapter) DeleteDB(config bool, force bool) error {
-
-	path, filter := a.partitionMngr.GetHead().GetTablePath()
-	a.logger.Info("Delete partition %s", path)
-	err := utils.DeleteTable(a.container, path, filter, a.cfg.QryWorkers)
-	if err != nil && !force {
-		return err
+func (a *V3ioAdapter) DeleteDB(configExists bool, force bool, fromTime int64, toTime int64) error {
+	//part.startTime+part.partitionInterval < maxt
```
imakhlin: delete died code

pkg/tsdb/v3iotsdb.go
```diff
@@ -182,31 +172,34 @@ func (a *V3ioAdapter) Querier(_ context.Context, mint, maxt int64) (*querier.V3i
...
 }
 
-func (a *V3ioAdapter) DeleteDB(config bool, force bool) error {
-
-	path, filter := a.partitionMngr.GetHead().GetTablePath()
-	a.logger.Info("Delete partition %s", path)
-	err := utils.DeleteTable(a.container, path, filter, a.cfg.QryWorkers)
-	if err != nil && !force {
-		return err
+func (a *V3ioAdapter) DeleteDB(configExists bool, force bool, fromTime int64, toTime int64) error {
```
imakhlin: IMHO, it is better to provide an input struct with default values instead of separate params.

pkg/utils/asynciter.go
```diff
@@ -49,7 +49,9 @@ type AsyncItemsCursor struct {
 	Cnt           int
 }
 
-func NewAsyncItemsCursor(container *v3io.Container, input *v3io.GetItemsInput, workers int) (*AsyncItemsCursor, error) {
+func NewAsyncItemsCursor(
+	container *v3io.Container, input *v3io.GetItemsInput,
+	workers int, shardingKeys []string) (*AsyncItemsCursor, error) {
```
imakhlin: I believe that number of workers should be taken from the configuration and configuration should be read-only singleton  accessible from everywhere in the code of this library

pkg/utils/timeutils.go
```diff
@@ -44,6 +44,8 @@ func Str2duration(duration string) (int64, error) {
 				multiply = 24 * 3600 * 1000
 			case "w":
 				multiply = 7 * 24 * 3600 * 1000
+			case "M":
+				multiply = 30 * 24 * 3600 * 1000
```
imakhlin: What about months/years with different number of days? Should be resolved by clendar

pkg/partmgr/partmgr.go
gtopper: Actually, we ought to be using `path.Join`. `filepath.Join` is system-dependent. As to why not using `Sprintf` instead, I would argue that there are minor advantages in terms of readability (can read left-to-right without looking for substitutions, purpose as a path is clear, no need to consider escape sequences).

pkg/config/config.go
gtopper: Also, if we're following [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601), `H` should be `h`.

pkg/config/config.go
gtopper: As noted in another comment, the correct function we ought to be using is actually `path.Join`.

pkg/partmgr/partmgr.go
imakhlin: on delete partition it should update the head anyway

pkg/partmgr/partmgr.go
imakhlin: 1. OS independent separator
2. it adds a Separator if necessary
3. Join calls Clean (https://golang.org/pkg/path/filepath/#Clean) on the result; in particular, all empty strings are ignored. On Windows, the result is a UNC path if and only if the first path element is a UNC path.

pkg/querier/seriesset.go
imakhlin: ok, but it has not been used yet.
I would suggest more human readable name, maybe `_in_order` = [true|false] (or `_sorted`)

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
 	"sync"
 )
 
-// Create new Partition Manager, for now confined to one Cyclic partition
-func NewPartitionMngr(cfg *config.DBPartConfig, path string) *PartitionManager {
-	newMngr := &PartitionManager{cfg: cfg, path: path, cyclic: true, ignoreWrap: true}
-	newMngr.headPartition = NewDBPartition(newMngr)
+// Create new Partition Manager
+func NewPartitionMngr(cfg *config.Schema, path string, cont *v3io.Container) *PartitionManager {
+	currentPartitionInterval, _ := utils.Str2duration(cfg.PartitionSchemaInfo.PartitionerInterval)
```
gtopper: Shouldn't we be checking the error return value?

pkg/partmgr/partmgr.go
gtopper: `path.Join(...)` 😄

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
+	newMngr := &PartitionManager{cfg: cfg, path: path, cyclic: false, container: cont, currentPartitionInterval: currentPartitionInterval}
+	for _, part := range cfg.Partitions {
+		path := fmt.Sprintf("%s/%d/", newMngr.path, part.StartTime/1000)
+		newPart := NewDBPartition(newMngr, part.StartTime, path)
+		newMngr.partitions = append(newMngr.partitions, newPart)
+		if newMngr.headPartition == nil {
+			newMngr.headPartition = newPart
+		} else {
+			if newMngr.headPartition.startTime < newPart.startTime {
```
gtopper: `else if`?

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
+	}
 	return newMngr
 }
 
 // Create and Init a new Partition
-func NewDBPartition(pmgr *PartitionManager) *DBPartition {
+func NewDBPartition(pmgr *PartitionManager, startTime int64, path string) *DBPartition {
+	rollupTime, _ := utils.Str2duration(pmgr.cfg.PartitionSchemaInfo.AggregatorsGranularity)
+	partitionInterval, _ := utils.Str2duration(pmgr.cfg.PartitionSchemaInfo.PartitionerInterval)
+	chunkInterval, _ := utils.Str2duration(pmgr.cfg.PartitionSchemaInfo.ChunckerInterval)
```
gtopper: Shouldn't we be checking the return error value?

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,201 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
 }
 
-func (p *PartitionManager) GetHead() *DBPartition {
+func (p *PartitionManager) updatePartitionInSchema(partition *DBPartition) error {
+	p.cfg.Partitions = append(p.cfg.Partitions, config.Partition{StartTime: partition.startTime, SchemaInfo: p.cfg.PartitionSchemaInfo})
+	data, err := json.Marshal(p.cfg)
+	if err != nil {
+		return errors.Wrap(err, "Failed to update new partition in schema file")
+	}
+	err = p.container.Sync.PutObject(&v3io.PutObjectInput{Path: p.path + config.SCHEMA_CONFIG, Body: data})
```
gtopper: Better use `path.Join`.

pkg/partmgr/partmgr.go
gtopper: Again, should be `path.Join`. `filepath.Join` is system-dependent... This is not a local path - should not depend on client OS.

pkg/querier/multipart.go
```diff
@@ -0,0 +1,217 @@
...
+func (it *mergedSeriesIterator) At() (t int64, v float64) {
+	return it.cur.At()
+}
+
+func (it *mergedSeriesIterator) Err() error {
+	return it.cur.Err()
+}
+
+// for future use, merge sort labels from multiple partitions
+func mergeLables(a, b []string) []string {
```
gtopper: This function appears to be unused.

pkg/tsdbctl/create.go
```diff
@@ -21,19 +21,30 @@ such restriction.
...
 	"github.com/spf13/cobra"
+	"github.com/v3io/v3io-tsdb/pkg/aggregate"
 	"github.com/v3io/v3io-tsdb/pkg/config"
 	"github.com/v3io/v3io-tsdb/pkg/tsdb"
+	"strconv"
+	"strings"
 )
 
+const SCHEMA_VERSION = 0
+const DEFAULT_STORAGE_CLASS = "local"
```
gtopper: Better replace this with idiomatic Go case.

pkg/tsdbctl/create.go
```diff
@@ -69,13 +82,70 @@ func (cc *createCommandeer) create() error {
...
+	if err := cc.validateFormat(cc.rollupInterval); err != nil {
+		return errors.Wrap(err, "Failed to parse rollup interval")
+	}
+
+	defaultRollup := config.Rollup{
+		Aggregators:            cc.defaultRollups,
+		AggregatorsGranularity: cc.rollupInterval,
+		StorageClass:           DEFAULT_STORAGE_CLASS,
+		SampleRetention:        cc.sampleRetention,
+		LayerRetentionTime:     "1y", //TODO
```
gtopper: TODO what?

pkg/querier/seriesset.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+
+		s.nullSeries = false
+
+		if s.aggrSeries.CanAggregate(s.partition.AggrType()) && s.maxt-s.mint > s.interval {
+
+			// create series from aggregation arrays (in DB) if the partition stored the desired aggregates
+			maxtUpdate := s.maxt
+			maxTime := s.iter.GetField("_maxtime")
+			if maxTime != nil && int64(maxTime.(int)) < s.maxt {
```
gtopper: We could add `GetFieldInt` to `ItemsCursor` and use that instead.

pkg/operators/mergers.go
```diff
@@ -0,0 +1,110 @@
...
+	currValids []bool
+	currIters  []ChildIterator
+	err        error
+}
+
+func NewIterSortMerger() *IterSortMerger {
+	newMerger := IterSortMerger{}
+	newMerger.iters = make([]ChildIterator, 0)
+	newMerger.done = make([]bool, 0)
+	newMerger.currValids = make([]bool, 0)
```
gtopper: These three lines do nothing afaict.

pkg/operators/mergers.go
```diff
@@ -0,0 +1,110 @@
...
+	return &newMerger
+}
+
+func (im *IterSortMerger) AppendIter(iter ChildIterator) {
+	im.iters = append(im.iters, iter)
+	im.done = append(im.done, false) //!iter.Next())
```
gtopper: Dead code ☠️

pkg/operators/mock.go
```diff
@@ -0,0 +1,97 @@
...
+implied. See the License for the specific language governing
+permissions and limitations under the License.
+
+In addition, you may not use the software for any purposes that are
+illegal under applicable law, and the grant of the foregoing license
+under the Apache 2.0 license is conditioned upon your compliance with
+such restriction.
+*/
+
+package operators
```
gtopper: Unused package.

pkg/aggregate/aggregate.go
```diff
@@ -58,11 +59,51 @@ var aggrToString = map[AggrType]string{
...
+	"rate":   {Name: "rate", Type: "array", Nullable: true, Items: "double"},
+	"stddev": {Name: "stddev", Type: "array", Nullable: true, Items: "double"},
+	"stdvar": {Name: "stdvar", Type: "array", Nullable: true, Items: "double"},
+}
+
+func SchemaFieldFromString(split []string, col string) ([]config.SchemaField, error) {
+	fieldList := make([]config.SchemaField, 0, len(split))
+	if len(split) != 0 && split[0] != "" {
+		for _, s := range split {
+			if strings.Compare(s, "*") == 0 {
```
gtopper: I think we can use string equality, no? :)

gtopper:
Superseded by #45.

PR #42, opened 2018-08-08
Partitions -- Fix skipping of start partition in query.

gtopper:


gtopper:
Approved by @talIguaz.

PR #43, opened 2018-08-08
Build paths independently from client OS.

gtopper:


imakhlin:
Sorry, but I didn't get your point.
In all cases where you did the change we do read the file (configuration) and it must be OS dependant path. This is not just a string in a model, but actual path that should be used to read the file in certain runtime OS

Look at the https://stackoverflow.com/questions/39181790/whats-the-difference-between-path-path-filepath-packages-in-go

gtopper:
You're right, I'll change config paths back. I meant to change just the non-local paths and forgot about v3io.yaml.

gtopper:
Actually none of them in dev are remote :)

PR #45, opened 2018-08-09
partitions PR fixes

talIguaz:


pkg/aggregate/aggregate.go
```diff
@@ -58,14 +59,56 @@ var aggrToString = map[AggrType]string{
...
+	"avg":    {Name: "avg", Type: "array", Nullable: true, Items: "double"},
+	"rate":   {Name: "rate", Type: "array", Nullable: true, Items: "double"},
+	"stddev": {Name: "stddev", Type: "array", Nullable: true, Items: "double"},
+	"stdvar": {Name: "stdvar", Type: "array", Nullable: true, Items: "double"},
+}
+
+func SchemaFieldFromString(aggregators []string, col string) ([]config.SchemaField, error) {
+	fieldList := make([]config.SchemaField, 0, len(aggregators))
+	if len(aggregators) != 0 && aggregators[0] != "" {
+		for _, s := range aggregators {
```
gtopper: I think it probably makes more sense to ignore empty strings throughout the loop.

pkg/aggregate/aggregate.go
```diff
@@ -58,14 +59,56 @@ var aggrToString = map[AggrType]string{
...
+			if trimmed == "*" {
+				fieldList = make([]config.SchemaField, 0, len(aggrToSchemaField))
+				for _, val := range aggrToSchemaField {
+					fieldList = append(fieldList, getAggrFullName(val, col))
+				}
+				return fieldList, nil
+			} else {
+				field, ok := aggrToSchemaField[trimmed]
+				if !ok {
+					return nil, fmt.Errorf("invalid aggragator type %s", trimmed)
```
imakhlin: please quote the '%s' so in case of "sum, ,avg" it will be clear what went wrong or/and just skip empty strings

pkg/config/config.go
```diff
@@ -54,42 +55,52 @@ type V3ioConfig struct {
...
-	// Metric specific policy
-	MetricsConfig map[string]MetricConfig `json:"metricsConfig,omitempty"`
+type Rollup struct {
+	Aggregators            string `json:"aggregators"`
+	AggregatorsGranularity string `json:"aggregatorsGranularity"`
+	//["cloud","local"] for the aggregators and sample chucks
+	StorageClass string `json:"storageClass"`
+	//in hours. 0  means no need to save samples
+	SampleRetention int `json:"sampleRetention"`
+	// format : 1m, 7d, 3h . Possible intervals: y/m/d/h/M
```
imakhlin: see https://en.wikipedia.org/wiki/ISO_8601

pkg/config/config.go
```diff
@@ -54,42 +55,52 @@ type V3ioConfig struct {
...
+	PartitionerInterval string   `json:"partitionerInterval"`
+	ChunckerInterval    string   `json:"chunckerInterval"`
+}
+
+type PartitionSchema struct {
+	Version                int      `json:"version"`
+	Aggregators            []string `json:"aggregators"`
+	AggregatorsGranularity string   `json:"aggregatorsGranularity"`
+	StorageClass           string   `json:"storageClass"`
+	SampleRetention        int      `json:"sampleRetention"`
```
imakhlin: isn't it duration? sometime it's `int` and sometimes `string`, why?

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,203 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
 	"sync"
 )
 
-// Create new Partition Manager, for now confined to one Cyclic partition
-func NewPartitionMngr(cfg *config.DBPartConfig, path string) *PartitionManager {
-	newMngr := &PartitionManager{cfg: cfg, path: path, cyclic: true, ignoreWrap: true}
-	newMngr.headPartition = NewDBPartition(newMngr)
+// Create new Partition Manager
+func NewPartitionMngr(cfg *config.Schema, partPath string, cont *v3io.Container) *PartitionManager {
+	currentPartitionInterval, _ := utils.Str2duration(cfg.PartitionSchemaInfo.PartitionerInterval)
```
imakhlin: do not ignore errors

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,203 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
+	newMngr := &PartitionManager{cfg: cfg, path: partPath, cyclic: false, container: cont, currentPartitionInterval: currentPartitionInterval}
+	for _, part := range cfg.Partitions {
+		partPath := path.Join(newMngr.path, strconv.FormatInt(part.StartTime/1000, 10)) + "/"
```
imakhlin: is this necessary to append "/" to the path?

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,203 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
+			newMngr.headPartition = newPart
+		}
+	}
 	return newMngr
 }
 
 // Create and Init a new Partition
-func NewDBPartition(pmgr *PartitionManager) *DBPartition {
+func NewDBPartition(pmgr *PartitionManager, startTime int64, path string) *DBPartition {
+	rollupTime, _ := utils.Str2duration(pmgr.cfg.PartitionSchemaInfo.AggregatorsGranularity)
```
imakhlin: same here, check for errors

pkg/partmgr/partmgr.go
```diff
@@ -18,123 +18,203 @@ under the Apache 2.0 license is conditioned upon your compliance with
...
-	defaultRollups aggregate.AggrType // Default Aggregation functions to apply on sample update
-	rollupTime     int64              // Time range per aggregation bucket
-	rollupBuckets  int                // Total number of buckets per partition
+	manager           *PartitionManager
+	path              string             // Full path (in the DB) to the partition
+	startTime         int64              // Start from time/date
+	partitionInterval int64              // Number of millis stored in the partition
+	chunkInterval     int64              // number of millis stored in each chunk
+	prefix            string             // Path prefix
+	retentionDays     int                // Keep samples for N hours
```
imakhlin: fix comment, why duration units are changing all over the way?
Note, string/int/ int64... if we decided to define duration in milliseconds, lets do it all the way. String should be used in configuration and CLI (human interfaces). internal representation should be in `millis` only

pkg/utils/timeutils.go
```diff
@@ -33,7 +33,7 @@ func Str2duration(duration string) (int64, error) {
 	multiply := 3600 * 1000 // hour by default
 	if len(duration) > 0 {
 		last := duration[len(duration)-1:]
-		if last == "m" || last == "h" || last == "d" || last == "y" {
+		if last == "m" || last == "h" || last == "d" || last == "y" || last == "M" || last == "w" {
```
imakhlin: is this compliant with ISO_8601?
see https://en.wikipedia.org/wiki/ISO_8601

imakhlin:
More tests needed, both unit and integration.

pkg/config/config.go
talIguaz: Agreed, as discussed let's fix it in another PR

pkg/partmgr/partmgr.go
talIguaz: Yes, because Nginx expectes URLs to end with a slash. it would be either save the path here with a trailing slash or change every Nginx call and append the slash

pkg/partmgr/partmgr.go
talIguaz: As said in other comment, will fix in a different PR

pkg/partmgr/partmgr.go
imakhlin: I prefer one place (close to HTTP request) that checks the URL and adds slash if missing

talIguaz: that way we'll need to wrap the ngnix calls or alternatively do this before every call which is more error prone.

pkg/utils/asynciter.go
```diff
@@ -139,12 +144,12 @@ func (ic *AsyncItemsCursor) NextItem() (v3io.Item, error) {
 	// Read response from channel
 	resp := <-ic.responseChan
 	if resp.Error != nil {
+		fmt.Println("error reading from response channel:", resp, "error", resp.Error, "request:", resp.Request().Input)
```
gtopper: Use logger?

pkg/config/config.go
```diff
@@ -62,7 +62,7 @@ type Rollup struct {
 	StorageClass string `json:"storageClass"`
 	//in hours. 0  means no need to save samples
 	SampleRetention int `json:"sampleRetention"`
-	// format : 1m, 7d, 3h . Possible intervals: y/m/d/h/M
+	// format : 1m, 7d, 3h . Possible intervals: m/d/h
```
gtopper: Can we have these in order?

pkg/tsdbctl/create.go
```diff
@@ -144,8 +144,8 @@ func (cc *createCommandeer) validateFormat(format string) error {
 		return fmt.Errorf("format is inncorrect, not a number")
 	}
 	unit := string(format[len(format)-1])
-	if !(unit == "y" || unit == "m" || unit == "d" || unit == "h" || unit == "M") {
-		return fmt.Errorf("format is inncorrect, not part of y,m,d,h,M")
+	if !(unit == "m" || unit == "d" || unit == "h") {
+		return fmt.Errorf("format is inncorrect, not part of m,d,h")
```
gtopper: Here too, I can't imagine why we would want this stuff to be out of order.

pkg/utils/timeutils.go
```diff
@@ -58,7 +52,7 @@ func Str2duration(duration string) (int64, error) {
 
 	i, err := strconv.Atoi(duration)
 	if err != nil {
-		return 0, errors.Wrap(err, "not a valid duration, use nn[s|h|m|d|y]")
+		return 0, errors.Wrap(err, "not a valid duration, use nn[s|h|m|d]")
```
gtopper: Here too, in order please. Also, this `nn[...]` notation is not very intuitive.

pkg/utils/timeutils.go
```diff
@@ -72,7 +66,7 @@ func Str2unixTime(tstr string) (int64, error) {
 	} else if strings.HasPrefix(tstr, "now-") {
 		t, err := Str2duration(tstr[4:])
 		if err != nil {
-			return 0, errors.Wrap(err, "not a valid time 'now-??', 'now' need to follow with nn[s|h|m|d|y]")
+			return 0, errors.Wrap(err, "not a valid time 'now-??', 'now' need to follow with nn[s|h|m|d]")
```
gtopper: In order please.

pkg/utils/labels.go
```diff
@@ -133,6 +133,24 @@ func (ls Labels) Hash() uint64 {
 	return hash.Sum64()
 }
 
+// Hash returns a hash value for the label set.
+func (ls Labels) LblHash() uint64 {
```
gtopper: This method is not in use.

pkg/utils/asynciter.go
gtopper: `ic.logger.Info(...)` (Or some other log level), although it would probably better to log this higher up the stack.

PR #46, opened 2018-08-09
Minor refactoring and unit test.

gtopper:


pkg/tsdbctl/add_test.go
```diff
@@ -0,0 +1,63 @@
+/*
```
talIguaz: please add a build tag

pkg/tsdbctl/add_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	suite.Require().Nil(err)
+	suite.Require().Equal(ts, []int64{1533814796000, 1533894796000, 1533899796000})
+	suite.Require().True(math.IsNaN(vs[0])) // NaN != NaN, so we have to check this explicitly
+	suite.Require().Equal(vs[1:], []float64{math.Inf(1), math.Inf(-1)})
+}
+
+func (suite *testSuite) TestStrToTVInvalidInput() {
+
+	ts, vs, err := strToTV("1533814796000,1533894796000,1533899796000", "1.2,5,z")
```
imakhlin: in this case you'll never test invalid time, since the test will break first on invalid value

pkg/tsdbctl/add_test.go
gtopper: Fixed.

pkg/tsdbctl/add_test.go
gtopper: No one said it checks for invalid time :).

imakhlin: There are two inputs, time and value. Why not to test them all for correctness?

gtopper: Fine fine

PR #50, opened 2018-08-13
Temp remove unsupported values

imakhlin:


pkg/aggregate/functions.go
```diff
@@ -138,8 +138,8 @@ func (a *LastAggregator) Aggregate(t int64, v float64) {
 }
 func (a *LastAggregator) UpdateExpr(col string, bucket int) string {
 	if math.IsNaN(a.val) {
```
gtopper: I don't get it... This can't happen anymore.

pkg/aggregate/functions.go
```diff
@@ -128,18 +132,19 @@ type LastAggregator struct {
...
 		a.val = v
 		a.lastT = t
 	}
 }
+
 func (a *LastAggregator) UpdateExpr(col string, bucket int) string {
 	if math.IsNaN(a.val) {
-		// TODO: replace with NaN when engine will support that syntax
-		return fmt.Sprintf("_%s_%s[%d]=%f;", col, a.attr, bucket, -math.MaxFloat64)
+		return ""
```
gtopper: Should remove this `if`. The condition will always evaluate as false.

pkg/aggregate/functions.go
gtopper: Apparently it still can... We still have `NaN` in many places.

pkg/aggregate/functions.go
```diff
@@ -22,6 +22,7 @@ package aggregate
 
 import (
 	"fmt"
+	"github.com/v3io/v3io-tsdb/pkg/utils"
```
gtopper: Better make this relative.

pkg/aggregate/aggregate_test.go
```diff
@@ -5,6 +5,7 @@ package aggregate
 import (
 	"fmt"
 	"github.com/stretchr/testify/assert"
+	"github.com/v3io/v3io-tsdb/pkg/utils"
```
gtopper: Better make this relative.

pkg/aggregate/functions.go
imakhlin: will do it in future PR that will be dealing with replacing NaN by `count` + `Inf` as agreed yesterday. For now I won't remove this safety check because engine does not support NaN yet and there is a scenario when `a.val == NaN`

PR #53, opened 2018-08-14
Replace nan with infinity

imakhlin:
- replace NaN in functions and aggregators by math.Inf(...)
- update init_array - use default aggregator's value


pkg/aggregate/functions.go
```diff
@@ -114,10 +115,10 @@ func (a *MinAggregator) UpdateExpr(col string, bucket int) string {
 // Maximum Aggregator
 type MaxAggregator struct{ FloatAggregator }
 
-func (a *MaxAggregator) Clear() { a.val = -math.MaxFloat64 } // TODO: use math.Inf(-1)
+func (a *MaxAggregator) Clear() { a.val = math.Inf(-1) }
 
 func (a *MaxAggregator) Aggregate(t int64, v float64) {
-	if !math.IsNaN(v) && (math.IsNaN(a.val) || v > a.val) {
+	if utils.IsDefined(v) && (utils.IsUndefined(a.val) || v > a.val) {
```
gtopper: You shouldn't have to defend from ±Inf - it should just work.

pkg/aggregate/functions.go
```diff
@@ -91,18 +92,18 @@ func (a *SumAggregator) Aggregate(t int64, v float64) {
 type SqrAggregator struct{ FloatAggregator }
 
 func (a *SqrAggregator) Aggregate(t int64, v float64) {
-	if !math.IsNaN(v) {
+	if utils.IsDefined(v) {
 		a.val += v * v
```
gtopper: Where would these nonnumerical values come from?

pkg/aggregate/functions.go
```diff
@@ -75,14 +75,15 @@ func (a *FloatAggregator) UpdateExpr(col string, bucket int) string {
 }
 
 func (a *FloatAggregator) InitExpr(col string, buckets int) string {
-	return fmt.Sprintf("_%s_%s=init_array(%d,'double');", col, a.attr, buckets)
+	// TODO: consider adding default init value per aggregator
+	return fmt.Sprintf("_%s_%s=init_array(%d,'double', %f);", col, a.attr, buckets, a.val)
```
gtopper: Let's be consistent on spacing.

pkg/aggregate/aggregate.go
```diff
@@ -132,13 +132,13 @@ func NewAggregatorList(aggrType AggrType) *AggregatorList {
...
-		list = append(list, &MinAggregator{FloatAggregator{attr: "min", val: math.MaxFloat64}}) // TODO: use math.Inf(1)
+		list = append(list, &MinAggregator{FloatAggregator{attr: "min", val: math.Inf(1)}})
 	}
 	if (aggrType & aggrTypeMax) != 0 {
-		list = append(list, &MaxAggregator{FloatAggregator{attr: "max", val: -math.MaxFloat64}}) // TODO: use math.Inf(-1)
+		list = append(list, &MaxAggregator{FloatAggregator{attr: "max", val: math.Inf(-1)}})
 	}
 	if (aggrType & aggrTypeLast) != 0 {
-		list = append(list, &LastAggregator{FloatAggregator{attr: "last"}, 0})
+		list = append(list, &LastAggregator{FloatAggregator{attr: "last", val: math.Inf(-1)}, 0})
```
gtopper: I don't think -Inf is really appropriate here. It shouldn't matter what we set it to, but still.

pkg/aggregate/functions.go
imakhlin: it's `go fmt ../../...` I have nothing to do with that

pkg/aggregate/functions.go
imakhlin: User input could contain NaN for example

pkg/aggregate/aggregate.go
imakhlin: Just using the same value as have been used at initialisation: `&LastAggregator{FloatAggregator{attr: "last", val: math.Inf(-1)}`

pkg/aggregate/functions.go
imakhlin: User input... NaN in CSV file for example

gtopper: Sure. But then it could also be Inf or -Inf, and it's incorrect to ignore those.

gtopper: Also, it would be legitimate for us to reject nonnumericals as input.

pkg/aggregate/functions.go
gtopper: I mean spacing within the update expression - that's not `go fmt`'s fault :)

pkg/aggregate/aggregate.go
gtopper: Still makes no sense, though.

pkg/aggregate/functions.go
imakhlin: Agree on rejecting invalid input. 
Regarding the +Inf/-Inf I see no reason to threat them as numeric values

pkg/aggregate/functions.go
imakhlin: it so negligible ...

PR #54, opened 2018-08-14
Fix sharding problem

talIguaz:
ignore 404 on `getItems` + remove commented code

### DIFF
--- a/pkg/utils/asynciter.go
+++ b/pkg/utils/asynciter.go
@@ -25,6 +25,7 @@ import (
 	"github.com/nuclio/logger"
 	"github.com/pkg/errors"
 	"github.com/v3io/v3io-go-http"
+	"strings"
 )
 
 type ItemsCursor interface {
@@ -73,7 +74,6 @@ func NewAsyncItemsCursor(
 				Path: input.Path, AttributeNames: input.AttributeNames, Filter: input.Filter,
 				ShardingKey: shardingKeys[i]}
 			_, err := container.GetItems(&input, 0, newAsyncItemsCursor.responseChan)
-
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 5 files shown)


pkg/utils/asynciter.go
```diff
@@ -143,6 +142,13 @@ func (ic *AsyncItemsCursor) NextItem() (v3io.Item, error) {
 
 	// Read response from channel
 	resp := <-ic.responseChan
+
+	// Ignore 404s
+	if resp.Error != nil && strings.Contains(resp.Error.Error(), "404") {
```
gtopper: This is pretty awful... Can't we get the actual http code from `resp`?

pkg/utils/asynciter.go
```diff
@@ -143,6 +142,13 @@ func (ic *AsyncItemsCursor) NextItem() (v3io.Item, error) {
...
+		fmt.Println("Got 404", resp.Error, "input:", resp.Request().Input)
```
gtopper: We really have to fix that logger...

pkg/utils/asynciter.go
imakhlin: Need to fill feature request to HTTP-GO lib to return error codes for each HTTP call
Please check with @pavius if it's something he can do quickly

pkg/utils/asynciter.go
imakhlin: Why not to use `ic.logger.Error(...` ?

talIguaz: the logger is not initialized there. so you need to either pass a logger every time you create an async iterator the problem is that one of the places doesn't have a logger to pass.
Either way our logging stuff isn't that good and Eran said he want to think what is the best practice and then notify us.. so we might as well wait with that a bit

pkg/utils/asynciter.go
talIguaz: He is too busy, we can open a PR with a fix by ourselves but not in this PR and i guess it's not *that* urgent

gtopper: Ok for now, but please add a TODO and see if you can restrict this condition further (so it doesn't match something like `1534324049`).

PR #60, opened 2018-08-16
TSDB-3

talIguaz:
substitute * with all aggregators on tsdb create


pkg/tsdbctl/create.go
```diff
@@ -94,8 +93,13 @@ func (cc *createCommandeer) create() error {
 		return errors.Wrap(err, "Failed to parse rollup interval")
 	}
 
+	rollups, err := aggregate.AggregatorsToStringList(cc.defaultRollups)
+	if err != nil {
+		return errors.Wrap(err, "Failed to parse default rollups")
```
gtopper: Don't capitalize error strings :)

PR #61, opened 2018-08-16
fix partition creation on historical data

dinal:


pkg/partmgr/partmgr.go
```diff
@@ -157,8 +161,19 @@ func (p *PartitionManager) createNewPartition(t int64) (*DBPartition, error) {
 		return nil, err
 	}
 	p.currentPartitionInterval = partition.partitionInterval
```
imakhlin: It would be better to separate creation logic from update current partition details

pkg/partmgr/partmgr.go
```diff
@@ -144,9 +144,13 @@ func (p *PartitionManager) TimeToPart(t int64) (*DBPartition, error) {
 					return p.partitions[i], nil
 				}
 			}
+			head := p.headPartition
+			part, _ := p.createNewPartition(p.currentPartitionInterval * (t / p.currentPartitionInterval))
```
imakhlin: if you'll separate `create` and `set` your code will be mush more readable

pkg/partmgr/partmgr.go
```diff
@@ -169,7 +184,9 @@ func (p *PartitionManager) updatePartitionInSchema(partition *DBPartition) error
 	if err != nil {
 		return errors.Wrap(err, "Failed to update new partition in schema file")
 	}
-	err = p.container.Sync.PutObject(&v3io.PutObjectInput{Path: path.Join(p.path, config.SCHEMA_CONFIG), Body: data})
+	if p.container != nil {
+		err = p.container.Sync.PutObject(&v3io.PutObjectInput{Path: path.Join(p.path, config.SCHEMA_CONFIG), Body: data})
+	}
```
imakhlin: else what?

pkg/partmgr/partmgr_test.go
```diff
@@ -0,0 +1,88 @@
...
+	part, _ = manager.TimeToPart(startTime + (interval * 2))
+	assert.Equal(tst, 3, len(manager.partitions))
+	assert.Equal(tst, manager.partitions[1], part)
+	//add first
+	part, _ = manager.TimeToPart(startTime)
+	assert.Equal(tst, 4, len(manager.partitions))
+	assert.Equal(tst, manager.partitions[0], part)
+}
+
+func createSchema() *config.Schema {
```
imakhlin: reuse or modify  pkg/tsdb/tsdbtest/tsdbtest.go:CreateSchema

dinal: i can't, it complainד about cyclic import

pkg/partmgr/partmgr.go
dinal: else nothing, file will not be updated (it can be nil only in tests)

imakhlin: it seems to me that `nil` container is an error rather than valid state

pkg/partmgr/partmgr_test.go
imakhlin: this is no excuse for code duplication - please refactor if necessary

PR #63, opened 2018-08-19
add validations for  metric and label names

imakhlin:
1. validate metric name
2. validate label name

### DIFF
--- a/pkg/tsdbctl/add.go
+++ b/pkg/tsdbctl/add.go
@@ -254,7 +254,7 @@ func strToLabels(name, lbls string) (utils.Labels, error) {
 			}
 
 			if err := utils.IsValidLabelName(splitLbl[0]); err != nil {
-				return nil, errors.Wrap(err, fmt.Sprintf("illegal lable name: '%s'", splitLbl[0]))
+				return nil, errors.Wrap(err, fmt.Sprintf("illegal label name: '%s'", splitLbl[0]))
 			}
 			lset = append(lset, utils.Label{Name: splitLbl[0], Value: splitLbl[1]})
 		}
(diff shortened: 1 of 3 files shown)


pkg/utils/validators.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+const (
+	metricNameValidationRegexStr = `^[a-zA-Z_:]([a-zA-Z0-9_:])*$`
+	labelValidationRegexStr      = `^[a-zA-Z_]([a-zA-Z0-9_])*$`
```
gtopper: Any particular reason we support colon in name but not label?

pkg/tsdbctl/add.go
```diff
@@ -248,6 +252,10 @@ func strToLabels(name, lbls string) (utils.Labels, error) {
 			if len(splitLbl) != 2 {
 				return nil, errors.New("labels must be in the form: key1=label1,key2=label2,...")
 			}
+
+			if err := utils.IsValidLabelName(splitLbl[0]); err != nil {
+				return nil, errors.Wrap(err, fmt.Sprintf("illegal lable name: '%s'", splitLbl[0]))
```
gtopper: Typo. Do we really need to wrap it anyway? There's no information here that's not already in `err`.

imakhlin: Wrapping it to create better stack trace.

pkg/utils/validators.go
imakhlin: No, this is what Prometheus do. We just mimic same convention.
see: https://github.com/prometheus/client_java/issues/28

PR #67, opened 2018-08-19
Avoid go get before build.

gtopper:
This causes releases to go get per deliverable, which is undesirable both in terms of build time and consistency across platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,15 +18,15 @@ get:
 	go get -v -t -tags "unit integration" $(TOPLEVEL_DIRS)
 
 .PHONY: test
-test: get
+test:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -18,15 +18,15 @@ get:
...
 	go test -race -tags unit -count 1 $(TOPLEVEL_DIRS)
 
 .PHONY: integration
-integration: get
+integration:
 	go test -race -tags integration -p 1 -count 1 $(TOPLEVEL_DIRS) # p=1 to force Go to run pkg tests serially.
 
 .PHONY: build
-build: get
+build:
```
talIguaz: Are you sure we dont want to `go get` before build?

PR #68, opened 2018-08-19
Fix prometheus query failure

pavius:
3 fixes to asynciter:

1. When an iterator received `Last = False`, it would call `GetItems` with a newly constructed `GetItemsInput` which did not have the `ShardingKey` set. This PR passes the originally constructed `GetItemsInput` to subsequent calls of `GetItems`, only setting `Marker` (this caused the issue)
2. `response` was released prior to being used. Now released with `defer`
3. Improper error was wrapped in case subsequent `GetItems` is called

DO NOT MERGE until this is verified with latest.


imakhlin:
The change looks fine to me.

pkg/utils/asynciter.go
```diff
@@ -156,22 +166,23 @@ func (ic *AsyncItemsCursor) NextItem() (v3io.Item, error) {
 	}
 
 	getItemsResp := resp.Output.(*v3io.GetItemsOutput)
-	shard := resp.Context.(int)
-	resp.Release()
+	defer resp.Release()
```
gtopper: Shouldn't this statement appear earlier in the flow? It seems to me like a leak is possible (e.g. if we get 404 and recurse several lines earlier. Not a regression here, but could be another issue.

PR #73, opened 2018-08-20
TSDB-7: Remove redundant check and add necessary validations.

imakhlin:
Note, all arrays initialised with 'zero' or '+-Inf', therefore there is no need to maintain initialisation state

### DIFF
--- a/pkg/aggregate/iterator.go
+++ b/pkg/aggregate/iterator.go
@@ -238,17 +238,10 @@ func isValidCell(cellIndex int, aSet *AggregateSet) bool {
 		cellIndex < aSet.length
 }
 
-func isValidAggCell(cellIndex int, aSet *AggregateSet, aType AggrType) bool {
-	return cellIndex >= 0 &&
-		cellIndex < len(aSet.dataArrays[aType]) &&
-		// cellIndex < aSet.maxCell &&
-		cellIndex < aSet.length
-}
-
 // function specific aggregation
 func (as *AggregateSet) updateCell(aggr AggrType, cell int, val float64) {
 
-	if !isValidAggCell(cell, as, aggr) {
+	if !isValidCell(cell, as) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/aggregate/iterator.go
```diff
@@ -267,15 +268,27 @@ func (as *AggregateSet) updateCell(aggr AggrType, cell int, val float64) {
 // return the value per aggregate or complex function
 func (as *AggregateSet) GetCellValue(aggr AggrType, cell int) (float64, bool) {
 
-	if cell > as.maxCell || cell >= as.length || !as.validCells[cell] { // TODO: should >Len return NaN or Zero ?
+	if !isValidCell(cell, as) {
+		return math.NaN(), false
+	}
+
+	dependsOnSum := aggr == aggrTypeStddev || aggr == aggrTypeStdvar
+	dependsOnCount := dependsOnSum || aggr == aggrTypeAvg
```
talIguaz: Isn't Avg also depends on sum?

pkg/aggregate/iterator.go
```diff
@@ -267,15 +268,27 @@ func (as *AggregateSet) updateCell(aggr AggrType, cell int, val float64) {
 // return the value per aggregate or complex function
 func (as *AggregateSet) GetCellValue(aggr AggrType, cell int) (float64, bool) {
 
-	if cell > as.maxCell || cell >= as.length || !as.validCells[cell] { // TODO: should >Len return NaN or Zero ?
+	if !isValidCell(cell, as) {
```
talIguaz: you do not check `cell > as.maxCell ` is that by design?

imakhlin: I see no reason for this check since maxCell is updated by adding a new cell

pkg/aggregate/iterator.go
imakhlin: yes it does. Why?

talIguaz: Then you should probably add `|| aggr == aggrTypeAvg` to the `dependsOnSum` because otherwise you might want the avg aggregator but the Sum aggregator is faulted so you will have corruption or panic

PR #75, opened 2018-08-21
TSDB-13: fix time to bucket calc

dinal:


gshatz:
@dinal did you noticed it is open for master?

PR #81, opened 2018-08-26
Unify v3io configuration environment variable. Update tests

imakhlin:
* Unify environment variable name
* Make test agnostic to runtime environment
* Print configuration in case of test failure

Example command for running tests:
`V3IO_CONF=<real-path>/v3io.yaml go test -tags "integration unit" ../../...`

Important Note: After merging this PR only `V3IO_CONF` will be respected

### DIFF
--- a/pkg/tsdb/tsdbtest/config_test.go
+++ b/pkg/tsdb/tsdbtest/config_test.go
@@ -32,8 +32,11 @@ func TestGetV3ioConfigPath(t *testing.T) {
 				// Make this test agnostic to environment variables at runtime (store & recover on exit)
 				configPathEnv := os.Getenv(config.V3ioConfigEnvironmentVariable)
 				os.Unsetenv(config.V3ioConfigEnvironmentVariable)
+
 				if _, err := os.Stat(filepath.Join(TsdbDefaultTestConfigPath, config.DefaultConfigurationFileName)); !os.IsNotExist(err) {
-					return func() {}
+					return func() {
+						os.Setenv(config.V3ioConfigEnvironmentVariable, configPathEnv)
+					}
 				} else {
 					path := TsdbDefaultTestConfigPath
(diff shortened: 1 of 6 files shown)


pkg/tsdb/tsdbtest/config_test.go
```diff
@@ -60,9 +68,9 @@ func TestGetV3ioConfigPath(t *testing.T) {
 		{description: "get config from env var",
 			expectedPath: config.DefaultConfigurationFileName,
 			setup: func() func() {
-				os.Setenv(V3ioConfigEnvironmentVariable, config.DefaultConfigurationFileName)
+				os.Setenv(config.V3ioConfigEnvironmentVariable, config.DefaultConfigurationFileName)
```
talIguaz: Here you are overriding the "maybe existing" config path, dont think you want to do that.

pkg/tsdb/tsdbtest/config_test.go
```diff
@@ -29,6 +29,9 @@ func TestGetV3ioConfigPath(t *testing.T) {
 		{description: "get config from package testdata",
 			expectedPath: filepath.Join(TsdbDefaultTestConfigPath, config.DefaultConfigurationFileName),
 			setup: func() func() {
+				// Make this test agnostic to environment variables at runtime (store & recover on exit)
```
talIguaz: This is a unit test. Unit tests don't really use the config since.. well.. they are unit tests so this change might be redundant.

pkg/tsdb/tsdbtest/config_test.go
imakhlin: It just recovers it's original value. In case I run the following command:
`V3IO_CONF=<real-path>/v3io.yaml go test -tags "integration unit" ../../...`
This test should ignore the env but recover the state afterwards

PR #91, opened 2018-08-28
Metrics

imakhlin:
- introduce performance metrics API
- changes to configuration loader (singleton + default, etc.)
- update tests


vendor/github.com/rcrowley/go-metrics/LICENSE
```diff
@@ -0,0 +1,29 @@
+Copyright 2012 Richard Crowley. All rights reserved.
```
dinal: did you mean to add this file?

pkg/appender/store.go
```diff
@@ -191,13 +200,26 @@ func (cs *chunkStore) ProcessGetResp(mc *MetricsCache, metric *MetricState, resp
 
 // Append data to the right chunk and table based on the time and state
 func (cs *chunkStore) Append(t int64, v interface{}) {
+	metricReporter, err := performance.DefaultReporterInstance()
```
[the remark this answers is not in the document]
imakhlin: because in context of this method I can't access the configuration.

pkg/partmgr/partmgr.go
```diff
@@ -170,15 +171,32 @@ func (p *PartitionManager) createAndUpdatePartition(t int64) (*DBPartition, erro
 	return partition, err
 }
 
-func (p *PartitionManager) updateSchema() error {
-	data, err := json.Marshal(p.cfg)
+func (p *PartitionManager) updateSchema() (err error) {
+	metricReporter, err := performance.DefaultReporterInstance()
 	if err != nil {
-		return errors.Wrap(err, "Failed to update new partition in schema file")
+		err = errors.Wrap(err, "unable to initialize performance metrics reporter")
```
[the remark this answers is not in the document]
imakhlin: I made this change to allow Timer to get result from the inner function, since timer signature does not allow return values I have to set the error within the inner function.

pkg/partmgr/partmgr_test.go
```diff
@@ -33,20 +33,40 @@ func TestCreateNewPartition(tst *testing.T) {
 	manager, _ := NewPartitionMngr(&schema, "/", nil)
 	interval := manager.currentPartitionInterval
 	startTime := interval + 1
+
 	//first partition
-	part, _ := manager.TimeToPart(startTime + interval)
+	part, err := manager.TimeToPart(startTime + interval)
```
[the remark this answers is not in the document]
imakhlin: added this check to catch configuration errors that might happen during the invocation of `TimeToPart`

vendor/github.com/rcrowley/go-metrics/LICENSE
imakhlin: IMHO, third party libraries should be vendorised "AS-IS"

pkg/appender/store.go
```diff
@@ -243,136 +265,155 @@ func (cs *chunkStore) chunkByTime(t int64) *attrAppender {
 }
 
 // write all pending samples to DB chunks and aggregators
-func (cs *chunkStore) WriteChunks(mc *MetricsCache, metric *MetricState) (bool, error) {
+func (cs *chunkStore) writeChunks(mc *MetricsCache, metric *MetricState) (hasPendingUpdates bool, err error) {
 
-	// return if there are no pending updates
-	if len(cs.pending) == 0 {
-		return false, nil
+	metricReporter, err := performance.DefaultReporterInstance()
```
dinal: you can use here ReporterInstanceFromConfig instead (from mc)

pkg/partmgr/partmgr.go
```diff
@@ -170,15 +171,32 @@ func (p *PartitionManager) createAndUpdatePartition(t int64) (*DBPartition, erro
 	return partition, err
 }
```
dinal: you can create a metrics reporter instance from config at v3iotsdb.go:133 add pass it to partmgr constructor thus avoiding calling DefaultReporterInstance each time and also using the right config

PR #94, opened 2018-09-02
Fixed missing aggregate data near partition end.

gtopper:


imakhlin:
Please add integration test

PR #97, opened 2018-09-02
Switch to single worker

imakhlin:
1. Use single worker by default
2. Fix issue with default timeout

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -35,7 +35,7 @@ const (
 	defaultNumberOfWorkers      = 1
 	defaultNumberOfQueryWorkers = 8
 	defaultBatchSize            = 64
-	defaultTimeout              = int(24 * time.Hour)
+	defaultTimeout              = 24 * time.Hour
 )
 
 type V3ioConfig struct {
@@ -179,6 +179,6 @@ func InitDefaults(cfg *V3ioConfig) {
 	}
 
 	if cfg.DefaultTimeout == 0 {
-		cfg.DefaultTimeout = defaultTimeout
+		cfg.DefaultTimeout = int(defaultTimeout)
 	}
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -31,10 +31,10 @@ const (
 	DefaultConfigurationFileName  = "v3io.yaml"
 	SCHEMA_CONFIG                 = ".schema"
 
-	defaultNumberOfWorkers      = 1
-	defaultNumberOfQueryWorkers = 8
-	defaultBatchSize            = 64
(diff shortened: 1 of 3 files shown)


pkg/tsdb/testdata/v3io.yaml.template
```diff
@@ -15,7 +15,5 @@ verbose: "warn"
 # Default timeout 10 seconds
 timeout: 10
 
-workers: 8
```
gtopper: Why delete it from the template, though?

imakhlin: no workers means default, i.e workers=1

pkg/config/config.go
```diff
@@ -24,11 +24,19 @@ import (
...
-const SCHEMA_CONFIG = ".schema"
+const (
+	V3ioConfigEnvironmentVariable = "V3IO_CONF"
+	DefaultConfigurationFileName  = "v3io.yaml"
+	SCHEMA_CONFIG                 = ".schema"
+
+	defaultNumberOfWorkers      = 1
+	defaultNumberOfQueryWorkers = 8
+	defaultBatchSize            = 64
+	defaultTimeout              = 24 * time.Hour
```
talIguaz: Now the units are in nanosecond instead of seconds. should change documentation in `v3io.yaml.template` and in `config.go`. Or change it here to seconds.

pkg/config/config.go
```diff
@@ -24,11 +24,19 @@ import (
...
 
-const V3ioConfigEnvironmentVariable = "V3IO_CONF"
-const DefaultConfigurationFileName = "v3io.yaml"
-const SCHEMA_CONFIG = ".schema"
+const (
+	V3ioConfigEnvironmentVariable = "V3IO_CONF"
+	DefaultConfigurationFileName  = "v3io.yaml"
+	SCHEMA_CONFIG                 = ".schema"
+
+	defaultNumberOfWorkers      = 1
```
gshatz: better to use defaultNumberOfIngestWorkers

pkg/tsdb/testdata/v3io.yaml.template
gshatz: I would leave it commented out so users will see the option to control it

pkg/config/config.go
```diff
@@ -35,7 +34,7 @@ const (
 	defaultNumberOfWorkers      = 1
 	defaultNumberOfQueryWorkers = 8
 	defaultBatchSize            = 64
-	defaultTimeout              = 24 * time.Hour
+	defaultTimeoutInSeconds     = 24 * 60 * 60 // 24 hours
```
talIguaz: consider using `time.Hour / time.Seconds` which will give you one hour in seconds, might look better than literal numbers

PR #99, opened 2018-09-03
TSDB-20: validate min max in query

dinal:


pkg/tsdb/v3iotsdb.go
```diff
@@ -173,6 +173,9 @@ func (a *V3ioAdapter) Close() error {
 
 // create a querier interface, used for time series queries
 func (a *V3ioAdapter) Querier(_ context.Context, mint, maxt int64) (*querier.V3ioQuerier, error) {
+	if maxt < mint {
+		return nil, errors.Errorf("Start time is lower than end time")
```
gshatz: please add the maxt & mint to the error print

PR #101, opened 2018-09-03
Tsdb 20 - fix test

dinal:


pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -135,6 +135,7 @@ func TestQueryData(t *testing.T) {
 		to           int64
 		expected     map[string][]tsdbtest.DataPoint
 		ignoreReason string
+		expectFail   bool
 	}{
 		{desc: "Should ingest and query one data point", metricName: "cpu",
```
imakhlin: Update description. This test suppose to fail, not to add  a data point.

PR #102, opened 2018-09-03
Clear returned error in Add and AddFast.

gtopper:
The purpose is to allow further writing to an object that encountered an error.


imakhlin:
Looks legitimate, but please clarify what problem it solves?

gtopper:
Updated description.

gtopper:
This is probably not enough, and needs some testing, so closing for now.

PR #103, opened 2018-09-03
Add validation for rollup interval and query step

talIguaz:


pkg/tsdbctl/create.go
```diff
@@ -105,6 +106,12 @@ func (cc *createCommandeer) create() error {
 		return errors.Wrap(err, "failed to calculate chunk interval")
 	}
 
+	chunk, _ := utils.Str2duration(chunkInterval)
+	rollupInterval, _ := utils.Str2duration(cc.rollupInterval)
```
gtopper: What about the errors?

pkg/tsdbctl/query.go
```diff
@@ -101,6 +101,12 @@ func (qc *queryCommandeer) query() error {
 		return err
 	}
 
+	chunkIntervalStr := qc.rootCommandeer.adapter.GetSchema().TableSchemaInfo.ChunckerInterval
+	chunk, _ := utils.Str2duration(chunkIntervalStr)
```
gtopper: What about the error?

pkg/tsdbctl/query.go
```diff
@@ -101,6 +101,10 @@ func (qc *queryCommandeer) query() error {
 		return err
 	}
 
+	if oneHourMillis%step != 0 && step%oneHourMillis != 0 {
+		return errors.New("interval step should be a divisor or a dividend of 1 hour. Example: 10m, 30m, 2h, etc..")
+	}
```
gtopper: We discussed removing this bit in today's status meeting, since we have to support such queries, and @yaronha has a fix for that.

pkg/tsdbctl/create.go
```diff
@@ -178,9 +197,26 @@ func (cc *createCommandeer) calculatePartitionAndChunkInterval(rateInHours int)
...
 }
 
+func (cc *createCommandeer) enforcePartitionInterval(partitionIntervalInHours, chunkIntervalInHours int) (int, error) {
+	rollupIntervalMillis, err := utils.Str2duration(cc.rollupInterval)
+	rollupIntervalHours := int(rollupIntervalMillis / int64(time.Hour/time.Millisecond))
+	if err != nil {
+		return 0, err
+	}
+	if rollupIntervalHours < 1 {
+		return partitionIntervalInHours, nil
```
gtopper: If rollup interval is under an hour we change nothing?

pkg/tsdbctl/create.go
```diff
@@ -212,3 +248,10 @@ func rateToHours(sampleRate string) (int, error) {
 		return 0, parsingError
 	}
 }
+
+func getNextNumberEvenlyDivided(n, d1, d2 int) int {
+	var i int
+	for i = n; i%d1 != 0 || i%d2 != 0; i-- {
+	}
```
gtopper: Why are we confident that this will always yield a desirable result? It seems rather brute force, with potential to run forever on certain inputs.

PR #104, opened 2018-09-03
Fixed comment.

gtopper:


pkg/config/config.go
```diff
@@ -61,7 +61,7 @@ type V3ioConfig struct {
 	DefaultTimeoutInSeconds int `json:"timeout,omitempty"`
 	// The size of batch to use during ingestion
 	BatchSize int `json:"batchSize,omitempty"`
-	// Size of sample in bytes for worst an best compression scenarios
+	// Sample size in bytes in worst compression scenario
```
imakhlin: I think it worth mention the recommended range of values

gtopper: Only fixing wrong comment in this PR, not adding new stuff.

PR #107, opened 2018-09-04
aggr-clean

yaronha:
fix aggregator queries in multiple parts and add integration test

### DIFF
--- a/pkg/aggregate/iterator.go
+++ b/pkg/aggregate/iterator.go
@@ -69,8 +69,9 @@ func (as *AggregateSeries) CanAggregate(partitionAggr AggrType) bool {
 	// keep only real aggregators
 	aggrMask := 0x7f & as.aggrMask
 	// make sure the DB has all the aggregators we need (on bits in the mask)
-	// and that the aggregator resolution is greater/eq to requested interval
-	return ((aggrMask & partitionAggr) == aggrMask) && as.interval >= as.rollupTime
+	// and that the requested interval is greater/eq to aggregator resolution and is an even divisor
+	return ((aggrMask & partitionAggr) == aggrMask) &&
+		as.interval >= as.rollupTime && (as.interval%as.rollupTime == 0)
 }
 
 func (as *AggregateSeries) GetAggrMask() AggrType {
@@ -176,6 +177,11 @@ func (as *AggregateSeries) NewSetFromChunks(length int) *AggregateSet {
 	for _, aggr := range rawAggregators {
 		if aggr&as.aggrMask != 0 {
 			dataArrays[aggr] = make([]float64, length, length) // TODO: len/capacity & reuse (pool)
+			if aggr == aggrTypeMax || aggr == aggrTypeMin || aggr == aggrTypeLast {
+				for i := 0; i < length; i++ {
+					dataArrays[aggr][i] = math.NaN()
+				}
+			}
 		}
 	}
 
@@ -256,11 +262,11 @@ func (as *AggregateSet) updateCell(aggr AggrType, cell int, val float64) {
 	case aggrTypeSqr:
 		as.dataArrays[aggr][cell] += val * val
 	case aggrTypeMin:
-		if val < as.dataArrays[aggr][cell] {
+		if math.IsNaN(as.dataArrays[aggr][cell]) || val < as.dataArrays[aggr][cell] {
 			as.dataArrays[aggr][cell] = val
 		}
 	case aggrTypeMax:
-		if val > as.dataArrays[aggr][cell] {
+		if math.IsNaN(as.dataArrays[aggr][cell]) || val > as.dataArrays[aggr][cell] {
 			as.dataArrays[aggr][cell] = val
 		}
 	case aggrTypeLast:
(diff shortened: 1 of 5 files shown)


pkg/aggregate/iterator.go
```diff
@@ -176,6 +177,11 @@ func (as *AggregateSeries) NewSetFromChunks(length int) *AggregateSet {
 	for _, aggr := range rawAggregators {
 		if aggr&as.aggrMask != 0 {
 			dataArrays[aggr] = make([]float64, length, length) // TODO: len/capacity & reuse (pool)
+			if aggr == aggrTypeMax || aggr == aggrTypeMin || aggr == aggrTypeLast {
+				for i := 0; i < length; i++ {
+					dataArrays[aggr][i] = math.NaN()
```
imakhlin: Note, we use to initialise the data arrays with `(+/-)Inf` rather than `NaN`

pkg/aggregate/iterator.go
```diff
@@ -256,11 +262,11 @@ func (as *AggregateSet) updateCell(aggr AggrType, cell int, val float64) {
 	case aggrTypeSqr:
 		as.dataArrays[aggr][cell] += val * val
 	case aggrTypeMin:
-		if val < as.dataArrays[aggr][cell] {
+		if math.IsNaN(as.dataArrays[aggr][cell]) || val < as.dataArrays[aggr][cell] {
```
imakhlin: If you would initialise with proper `Int` value you won't need to change this

pkg/querier/querier.go
```diff
@@ -106,9 +106,13 @@ func (q *V3ioQuerier) selectQry(name, functions string, step int64, windows []in
 func (q *V3ioQuerier) queryNumericPartition(
 	partition *partmgr.DBPartition, name, functions string, step int64, windows []int, filter string) (SeriesSet, error) {
 
-	mint, maxt := partition.GetPartitionRange(q.maxt)
+	mint, maxt := partition.GetPartitionRange()
 	if q.mint > mint {
 		mint = q.mint
+		if step != 0 {
+			// temporary Aggregation fix: if mint is not aligned with steps, move it to the next step tick
+			mint += (maxt - q.mint) % step
```
imakhlin: you can round up with `mint += step-mint%step`
and round down with `mint -= mint%step`

but I think it is expected that reminder interval should also be counted.

test/aggregates/aggregates_test.go
```diff
@@ -0,0 +1,208 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		index = (index + 1) % len(testConfig.Values)
+		total++
+	}
+	fmt.Println("total samples written:", total)
+
+	appender.WaitForCompletion(0)
+	time.Sleep(time.Second * 2)
```
imakhlin: Why sleep?

test/aggregates/aggregates_test.go
```diff
@@ -0,0 +1,208 @@
...
+			}
+
+		}
+		if iter.Err() != nil {
+			t.Fatal(iter.Err())
+		}
+
+		if aggr == "count" {
+			fmt.Println("\nTotal count read:", totalCount)
+			if totalCount != total {
```
imakhlin: use Go test frameworks like testify, see `github.com/stretchr/testify/assert`

test/aggregates/aggregates_test.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+type metricContext struct {
+	lset utils.Labels
+	ref  uint64
+}
+
+func TestAggregates(t *testing.T) {
+
+	testConfig := TestConfig{
```
imakhlin: I would recommend separating the test onto stages:
1. Setup - prepare the data and prerequisites required for running the test like table with desired partitioning, etc
2. TearDown - cleanup after test (could be skipped if test has failed to make investigation possible)
3. The test code itself.

Nice example: `pkg/aggregate/aggregate_test.go`

imakhlin: This integration test wont pass twice.

test/aggregates/aggregates_test.go
```diff
@@ -0,0 +1,208 @@
+// +build integration
```
gtopper: Should rename file accordingly (`aggregates_integration_test`).

pkg/aggregate/iterator.go
yaronha: @imakhlin as discussed i think its simpler/cleaner (not confined to expr constraints), u can change if to the other way if u think its critical

pkg/aggregate/iterator.go
yaronha: see above

pkg/querier/querier.go
yaronha: @imakhlin expected return is results with time > query mint, with your suggestion the response will not comply with the user request

test/aggregates/aggregates_test.go
```diff
@@ -0,0 +1,208 @@
...
+implied. See the License for the specific language governing
+permissions and limitations under the License.
+
+In addition, you may not use the software for any purposes that are
+illegal under applicable law, and the grant of the foregoing license
+under the Apache 2.0 license is conditioned upon your compliance with
+such restriction.
+*/
+
+package aggregates
```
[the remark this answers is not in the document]
yaronha: @imakhlin  ok, please place it in there and can also use the create/tear-down flow

gshatz:
I am merging this pr although there are still some changes to be done in the integration tests due to delivery constraints

PR #114, opened 2018-09-05
Remove default rate

gshatz:
with an informative error to add it in case it missing

### DIFF
--- a/pkg/tsdbctl/create.go
+++ b/pkg/tsdbctl/create.go
@@ -33,7 +33,7 @@ import (
 const (
 	schemaVersion               = 0
 	defaultStorageClass         = "local"
-	defaultIngestionRate        = "1/s"
+	defaultIngestionRate        = "-1"
 	defaultRollupInterval       = "1h"
 	defaultShardingBuckets      = 8
 	defaultSampleRetentionHours = 0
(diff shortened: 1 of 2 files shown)


pkg/tsdbctl/create.go
```diff
@@ -33,7 +33,7 @@ import (
 const (
 	schemaVersion               = 0
 	defaultStorageClass         = "local"
-	defaultIngestionRate        = "1/s"
+	defaultIngestionRate        = "-1"
```
gtopper: Just make it empty? `-1` would make sense if it were an integer.

pkg/tsdbctl/create.go
```diff
@@ -182,7 +182,10 @@ func (cc *createCommandeer) calculatePartitionAndChunkInterval(rateInHours int)
 }
 
 func rateToHours(sampleRate string) (int, error) {
-	parsingError := errors.New(`not a valid rate. Accepted pattern: [0-9]+/[hms]. Examples: 12/m`)
+	if sampleRate == "-1" {
+		return 0, errors.New(`sample rate not provided! Please provide sample rate with --rate flag in the format of [0-9]+/[hms]. Example: 12/m`)
```
gtopper: Better move this error one level up, or it will get parsed in a (false) parse error.

PR #117, opened 2018-09-08
Fix label values

pavius:
`LabelValues` should be able to return the values of a given label. Before this PR, `LabelValues` only supporting getting the `__name__` label, effectively returning metric names. This PR introduces support for returning values of other labels. It does so by getting all the `_lset`s in the first partition, parsing them and looking for the given label.

While this implementation is limited (as it only gets the label values of the label in the first partition), it is good enough for netops and other cases where all metrics are created at once.

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -168,7 +168,7 @@ func (q *V3ioQuerier) Close() error {
 
 func (q *V3ioQuerier) getMetricNames() ([]string, error) {
 	input := v3io.GetItemsInput{
-		Path: q.cfg.Path + "/names/",
+		Path:           q.cfg.Path + "/names/",
 		AttributeNames: []string{"__name"},
 	}
 
@@ -207,7 +207,7 @@ func (q *V3ioQuerier) getLabelValues(labelKey string) ([]string, error) {
 
 	// get all labelsets
 	input := v3io.GetItemsInput{
-		Path: q.partitionMngr.GetPartitionsPaths()[0],
+		Path:           q.partitionMngr.GetPartitionsPaths()[0],
 		AttributeNames: []string{"_lset"},
 	}
 
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -198,8 +198,10 @@ func (q *V3ioQuerier) getLabelValues(labelKey string) ([]string, error) {
 		return nil, err
 	}
 
+	partitionPaths := q.partitionMngr.GetPartitionsPaths()
+
 	// if no partitions yet - there are no labels
-	if len(q.partitionMngr.GetPartitionsPaths()) == 0 {
+	if len(partitionPaths) == 0 {
 		return nil, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/querier/querier.go
```diff
@@ -154,27 +154,97 @@ func (q *V3ioQuerier) queryNumericPartition(
...
+func (q *V3ioQuerier) getLabelValues(labelKey string) ([]string, error) {
+
+	// sync partition manager (hack)
+	err := q.partitionMngr.ReadAndUpdateSchema()
+	if err != nil {
+		return nil, err
+	}
+
+	// if no partitions yet - there are no labels
+	if len(q.partitionMngr.GetPartitionsPaths()) == 0 {
```
gtopper: You better keep the result in a local variable. Despite its name, this method is not a simple getter.

pkg/querier/querier.go
```diff
@@ -154,27 +154,97 @@ func (q *V3ioQuerier) queryNumericPartition(
...
+				labelValuesMap[splitLabel[1]] = struct{}{}
+			}
+		}
+	}
+
+	if iter.Err() != nil {
+		q.logger.InfoWith("Failed to read names, assume empty list", "err", iter.Err().Error())
+	}
+
+	labelValues := []string{}
```
gtopper: I think this is usually better defined as `var labelValues []string`?

pkg/querier/querier.go
```diff
@@ -154,27 +154,97 @@ func (q *V3ioQuerier) queryNumericPartition(
...
+			// if we got two elements and the first element (the key) is equal to what we're looking
+			// for, save the label value in the map. use a map to prevent duplications
+			if len(splitLabel) == 2 && splitLabel[0] == labelKey {
+				labelValuesMap[splitLabel[1]] = struct{}{}
+			}
+		}
+	}
+
+	if iter.Err() != nil {
+		q.logger.InfoWith("Failed to read names, assume empty list", "err", iter.Err().Error())
```
gtopper: You changed the text of the original log message, so I would assume you meant to use the new text here as well.

pkg/querier/querier.go
pavius: I don't understand why a local variable is needed. Behind the scenes the return value is kept on the stack, equivalent to a local variable

pkg/querier/querier.go
```diff
@@ -154,27 +154,99 @@ func (q *V3ioQuerier) queryNumericPartition(
...
+			// if we got two elements and the first element (the key) is equal to what we're looking
+			// for, save the label value in the map. use a map to prevent duplications
+			if len(splitLabel) == 2 && splitLabel[0] == labelKey {
+				labelValuesMap[splitLabel[1]] = struct{}{}
+			}
+		}
+	}
+
+	if iter.Err() != nil {
+		q.logger.InfoWith("Failed to read label values, assume empty list", "err", iter.Err().Error())
```
gtopper: Log is a bit inconsistent with change to original log message.

PR #124, opened 2018-09-20
Add compression ratio print to 'check' operation

gshatz:


pkg/tsdbctl/check.go
```diff
@@ -122,7 +122,13 @@ func (cc *checkCommandeer) check() error {
 				if iter.Err() != nil {
 					return errors.Wrap(iter.Err(), "failed to read iterator")
 				}
-				fmt.Printf("Total Size: %d, Count: %d\n", len(bytes), count)
+
+				compressionRatio := 0.0
+				if count > 0 {
+					compressionRatio = float64(len(bytes)) / float64(count)
```
imakhlin: * extract `bytesCount := len(bytes)` and reuse it
* format floating number, for example `%.2f`

imakhlin:
Try to make `check` great again ;-)

PR #128, opened 2018-09-22
Vendor

imakhlin:
* "vendorise" all dependencies
* Create `README.md` with revisions of the dependent packages


internal/pkg/performance/metrics.go
```diff
@@ -20,11 +21,13 @@ var once sync.Once
 const (
 	STDOUT = "stdout"
 	STDERR = "stderr"
+
+	notRuningState = 0
+	runningState   = 1
```
gtopper: There's stuff in this PR that doesn't seem to be related to vendorization?

imakhlin: reverted this change

pkg/appender/appender.go
```diff
@@ -38,8 +38,8 @@ const channelSize = 4048
 const maxSamplesBatchSize = 16
 const queueStallTime = 1 * time.Millisecond
 
-const minimalUnixTime = 20000000 // year 1970
-const maxUnixTime = 13569465600  // year 2400
+const minimalUnixTimeMs = 0          // year 1970
```
imakhlin: Note, changed minTime to `0` to allow integration tests to run as they should

PR #129, opened 2018-09-23
verify time is in a valid range (year 197-2400), avoid errors due to wrong units (e.g sec vs milisec)

yaronha:


pkg/appender/appender.go
```diff
@@ -215,20 +220,32 @@ func (mc *MetricsCache) Add(lset utils.LabelsIfc, t int64, v interface{}) (uint6
...
+	err = metric.error()
 	metric.setError(nil)
 
 	mc.appendTV(metric, t, v)
 
 	return err
 }
 
+func verifyTimeValid(t int64) error {
+	if t > 13569465600000 || t < 20000000000 {
```
imakhlin: consider using human readable time format for better readability
Example: https://stackoverflow.com/questions/20924303/date-time-comparison-in-golang

PR #130, opened 2018-09-23
add method to init the TSDB config from a struct (vs yaml/file), for v3io frames

yaronha:


pkg/config/config.go
```diff
@@ -173,6 +173,12 @@ func GetOrLoadFromData(data []byte) (*V3ioConfig, error) {
 	return instance, failure
 }
 
+// update defaults when using config struct
+func GetOrLoadFromStruct(cfg *V3ioConfig) error {
```
imakhlin: config instance is a singleton and should not be changed after initialisation.
`GetOrLoadFromStruct` suppose to return an existing or new configuration instance and optional error

PR #137, opened 2018-09-26
Align TSDB CLI with @sharonl recommendations

imakhlin:
II. CLI Flags
1. Global flags
    * Rename the config-file -c flag to -g (to allow using -c for the container name — see #1.b).
    * Rename the container-identifier -u flag to -c; (depends on #1.a).
    * Separate the username and password from the -s|--server flag (as already done in the configuration file): instead of setting the server flag to “<username>:<password>@<server URL>”, set it to “<server URL>” and add separate username (-u | --username) and password (-w | --password) flags; (depends on #1.b for using -u).
    * Verify that the default value of the -v|--verbose flag should be "debug", as currently set in the code (see #IV.1); the comment for the verbose configuration in the configuration template (pkg/tsdb/testdata/v3io.yaml.template) says that the default value is "info".
    * If the configuration-file flag (-c|--config / -g after #1.a is implemented) isn’t set, don’t return an error if the default configuration file (currently v3io.yaml in the current directory) can’t be loaded (as is currently the case).
If the configuration-file flag is set, require the file-path string (don’t rely on the default path).
2. create command
    * Remove the currently unsupported --sample-retention flag (Bug TSDB-31). --> DONE today (25.9.18). When this flag is returned, rename its -a CLI flag to -r; (depends on #2.b).
    * Rename the -r|--rollups flags to -a|--aggregates; (depends on #2.a).
    * Rename the --rollup-interval flag to --aggregation-granularity; (we can leave the short flag a -i).
3. add command
    * Do not externalize the --delay and --stdin flags; this request was approved f2f by Yaron H..
    * Add an ingest alias.
4. query command
    * Rename the --aggregators flag to --aggregates.
    * Rename the --step flag to --aggregation-interval; (we can leave the short flag a -i)
5. del command — the following changes were agreed upon in a f2f discussion with Golan and Igor:
    * Rename the -f|--force flag to -i|--ignore-errors.
    * Rename the -d|--del-config flag to -a|--all and change the behavior to ignore the -b|--begin and -e|--end flags and always delete the entire table (content & configuration); edit the description accordingly to “Delete the TSDB table (including all content and the configuration schema file)”.  (In the f2f discussion we spoke about renaming the flag to -d|-del-all, but in rethinking it I prefer the new alternative that I suggested here.)
    * Add -f|--force flags to force deletion of the table/partial content and change the behavior so that when this flag isn’t set the command displays an interactive prompt to verify the delete operation (“You are about to delete the <table name> table. Are you sure? (y/n)” or “You are about to delete content from the <table name> table. Are you sure? (y/n)” when the delete-all flag isn’t set); (depends on #5.a for using -f|--force). According to Golan, there might already be an open JIRA ticket to add this option.
6. time command — don’t externalize this command (approved by Golan, Igor, and Yaron H.).

### DIFF
--- a/pkg/tsdbctl/delete.go
+++ b/pkg/tsdbctl/delete.go
@@ -21,17 +21,21 @@ such restriction.
 package tsdbctl
 
 import (
+	"bufio"
 	"fmt"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/v3io/v3io-tsdb/pkg/utils"
+	"os"
+	"strings"
 	"time"
 )
 
 type delCommandeer struct {
 	cmd            *cobra.Command
 	rootCommandeer *RootCommandeer
-	delConfig      bool
+	deleteAll      bool
+	ignoreErrors   bool
 	force          bool
 	fromTime       string
 	toTime         string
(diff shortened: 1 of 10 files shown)


pkg/tsdbctl/delete.go
```diff
@@ -87,11 +92,45 @@ func (ic *delCommandeer) delete() error {
 			return err
 		}
 	}
-	err = ic.rootCommandeer.adapter.DeleteDB(ic.delConfig, ic.force, from, to)
+
+	if !ic.force {
+		confirmedByUser, err := getConfirmation(
+			fmt.Sprintf("You are about to delete the '%s' table. Are you sure?", ic.rootCommandeer.v3iocfg.Path))
```
dinal: add what you expect the user to enter "Are you sure? y/n"

pkg/tsdbctl/tsdbctl.go
```diff
@@ -161,10 +161,10 @@ func (rc *RootCommandeer) populateConfig(cfg *config.V3ioConfig) error {
 		if slash == -1 || len(rc.v3ioPath) <= slash+1 {
 			if rc.container != "" {
 				cfg.Container = rc.container
-				cfg.V3ioUrl = rc.v3ioPath
-			} else {
+			} else if cfg.Container == "" {
```
dinal: something is off here, cfg.Container will be empty for sure as it's not assigned anywhere before

pkg/tsdbctl/delete.go
Sharon-iguazio: I can confirm that when running the command, ` [y/n]` appears at the end of the question.

pkg/tsdbctl/tsdbctl.go
imakhlin: look at the signature `func (rc *RootCommandeer) populateConfig(cfg *config.V3ioConfig) error {`
`cfg` is input to this methos and it may come with all the settings loaded from some source

PR #139, opened 2018-09-26
support future times in Str2Unix function

yaronha:


pkg/utils/timeutils_test.go
```diff
@@ -66,6 +66,14 @@ func (suite *testTimeSuite) TestStr2durationBadFormat() {
 	suite.Require().Error(err)
 }
 
+func (suite *testTimeSuite) TestStr2unixTime() {
+	x, err := Str2unixTime("now+1m")
+	suite.Require().Nil(err)
+	y, err := Str2unixTime("now-1m")
+	suite.Require().Nil(err)
+	suite.Require().Equal(x-y, int64(120000))
```
imakhlin: equality check may not work properly since "now" is evaluated for each invocation

yaronha: ok, will allow few ms win

imakhlin:
Replaced with PR #145

PR #141, opened 2018-09-27
Update README & Refactor (Move create schema API from create.go to schema.go)

imakhlin:
* Update README 
* Add API for creating default schema programmatically, not just for testing
* Move create DB business logic from create command to schema.go
* Update tests accordingly


pkg/tsdb/schema/schema.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/pkg/errors"
+	"github.com/v3io/v3io-tsdb/pkg/aggregate"
+	"github.com/v3io/v3io-tsdb/pkg/config"
+	"strings"
+)
+
+const (
+	SchemaVersion                 = 0
+	DefaultStorageClass           = "local"
+	DefaultIngestionRate          = ""
```
talIguaz: It is not used. either delete it for now, or calculate the chunk&partition interval from the rate.

pkg/partmgr/partmgr_test.go
```diff
@@ -72,14 +72,18 @@ func TestCreateNewPartition(tst *testing.T) {
 }
 
 func getPartitionManager(tst *testing.T) *PartitionManager {
-	schema := testutils.CreateSchema(tst, "*")
+	schm, err := schema.NewSchema([]string{"*"})
```
talIguaz: You could just left it as is, and let the testutil.CreateSchema deal with errors and test fails.

pkg/tsdbctl/create.go
```diff
@@ -115,13 +103,13 @@ func (cc *createCommandeer) create() error {
 	defaultRollup := config.Rollup{
```
talIguaz: Why don't just use the `NewShema` here also?

pkg/tsdbctl/create.go
```diff
@@ -138,20 +126,20 @@ func (cc *createCommandeer) create() error {
...
 		AggregatorsGranularity: cc.rollupInterval,
-		StorageClass:           defaultStorageClass,
+		StorageClass:           schema.DefaultStorageClass,
 		SampleRetention:        cc.sampleRetention,
 		ChunckerInterval:       tableSchema.ChunckerInterval,
 		PartitionerInterval:    tableSchema.PartitionerInterval,
 	}
 
-	schema := config.Schema{
+	schm := config.Schema{
```
talIguaz: why don't just use the `NewSchema` here also?

pkg/tsdb/schema/schema.go
imakhlin: user MUST provide `IngestionRate` therefore NO DEFAULT value

pkg/tsdb/schema/schema.go
```diff
@@ -1,40 +1,66 @@
...
 	DefaultAggregationGranularity = "1h"
 	DefaultShardingBuckets        = 8
 	DefaultLayerRetentionTime     = "1y"
 	DefaultSampleRetentionTime    = 0
-	DefaultPartitionInterval      = "48h"
-	DefaultChunkInterval          = "1h"
 )
 
-func NewSchema(aggregates []string) (*config.Schema, error) {
+func NewSchema(sampleRate, aggregatorGranularity, aggregatesList string, minChunkSize, maxChunkSize, maxSampleSize, maxPartitionSize, sampleRetention, shardingBuckets int) (*config.Schema, error) {
```
talIguaz: I wouldn't expose the parameters for the rate calculation to the user. the way to change them is by config if you want. (`minChunkSize, maxChunkSize, maxSampleSize, maxPartitionSize`)

PR #144, opened 2018-09-30
Discard empty results

talIguaz:
When querying tsdb, if you have an empty step (meaning no data was received for that period) instead of returning NaN the iterator will just skip that step


pkg/aggregate/iterator.go
```diff
@@ -116,7 +116,8 @@ func (as *AggregateSeries) NewSetFromAttrs(
 	}
 
 	for _, aggr := range rawAggregators {
-		if aggr&as.aggrMask != 0 {
+		// Always have count aggregator by default
+		if aggr&as.aggrMask != 0 || aggr&aggrTypeCount != 0 {
```
imakhlin: aggr&aggrTypeCount != 0 will always be TRUE

pkg/aggregate/iterator.go
```diff
@@ -175,7 +176,8 @@ func (as *AggregateSeries) NewSetFromChunks(length int) *AggregateSet {
 	dataArrays := map[AggrType][]float64{}
 
 	for _, aggr := range rawAggregators {
-		if aggr&as.aggrMask != 0 {
+		// Always have count aggregator by default
+		if aggr&as.aggrMask != 0 || aggr&aggrTypeCount != 0 {
```
imakhlin: aggr&aggrTypeCount != 0 will always be TRUE

pkg/aggregate/iterator.go
talIguaz: Yes, it will always be TRUE once per loop. that way the count aggregator will be by default always as said in the comment above.

pkg/querier/series.go
```diff
@@ -257,10 +257,20 @@ func (s *aggrSeriesIterator) Seek(t int64) bool {
...
 // advance to the next time interval/bucket
 func (s *aggrSeriesIterator) Next() bool {
+
 	if s.index >= s.aggrSet.GetMaxCell() {
 		return false
 	}
 
+	nextIndex := s.index + 1
+	if nextIndex >= 0 && !s.aggrSet.DoesCellHaveData(nextIndex) {
+		for nextIndex >= 0 && nextIndex <= s.aggrSet.GetMaxCell() && !s.aggrSet.DoesCellHaveData(nextIndex) {
```
imakhlin: isn't is the same as
```
for nextIndex := s.index + 1; nextIndex <= s.aggrSet.GetMaxCell() && !s.aggrSet.DoesCellHaveData(nextIndex); nextIndex ++{
...
}
```

pkg/aggregate/aggregate.go
```diff
@@ -122,16 +122,22 @@ func AggregatorsToStringList(aggregators string) ([]string, error) {
...
 		if trimmed != "" {
 			aggr, ok := aggrTypeString[trimmed]
 			if !ok {
 				return aggrList, fmt.Errorf("invalid aggragator type '%s'", s)
 			}
+			hasAggregators = true
 			aggrList = aggrList | aggr
 		}
 	}
+	// Always have count aggregator by default
```
yaronha: @talIguaz  @igo isnt it simpler to just check if aggrList != 0 ?

talIguaz: the slice of strings here is created after calling `strings.split(<agg>, ',')` so if someone by mistake or whatever entered aggregators as let's say `  , ` , the slice will not be empty but there also will not be any aggregators

pkg/aggregate/iterator.go
yaronha: @talIguaz i think its better to just set aggrMask to include count (i.e. |= aggrTypeCount) in NewAggregateSeries assuming aggrMask!= 0  (line 54)

PR #145, opened 2018-09-30
Fix to yaron's PR - support future time string

imakhlin:
Replaces PR #139 

* Add missing functionality
* Fix unit test


yaronha:
LGTM

PR #146, opened 2018-09-30
TSDB-32: Do not append if sample is too old.

gtopper:


pkg/chunkenc/xor.go
```diff
@@ -163,6 +163,11 @@ func (a *xorAppender) Append(t int64, v float64) {
 	var tDelta uint64
 	num := *a.samples
 
+	// Do not append if sample is too old.
+	if t < a.t {
+		return
```
gshatz: please add a debug print for discarding this sample

gtopper: That's actually not possible without turning this PR into a logger refactoring PR.

pkg/chunkenc/chunkenc_test.go
```diff
@@ -127,38 +130,6 @@ func TestBstream(t *testing.T) {
 
 }
 
-func DecodeTest(blob string) error {
```
gshatz: why did you delete this test?

PR #147, opened 2018-09-30
Add config debug print

gshatz:


pkg/tsdb/v3iotsdb.go
```diff
@@ -146,6 +146,7 @@ func (a *V3ioAdapter) connect() error {
 	}
 
 	a.logger.Info("Starting V3IO TSDB client, server is at %s", fullpath)
+	a.logger.Debug("Running with configuration: %+v\n", a.cfg)
```
imakhlin: Output of this line is not user friendly and not useful.
You have to implement String() method for `V3ioConfig` and similar types in order to be able to print them out in human readable form

imakhlin: ok, I missed the `+` in `%+v`
LGTM

PR #149, opened 2018-10-02
Logger factory

imakhlin:
* simple and safe API for getting a logger from everywhere in the project


imakhlin:
@pavius recommended to pass parent logger everywhere in order to derive child logger from it and create hierarchical logs

PR #160, opened 2018-10-04
Support time.RFC3339Nano

tebeka:


pkg/utils/timeutils_test.go
```diff
@@ -106,6 +106,15 @@ func (suite *testTimeSuite) TestStr2unixTimeWithNowPlusMinus() {
 	suite.Require().Equal(expectedDuration, endTime-startTime)
 }
 
+func (suite *testTimeSuite) TestStr2unixTimeFRC() {
+	// RFC3339
+	_, err := Str2unixTime("2018-10-04T15:09:36+03:00")
```
gtopper: Might as well check the result.

pkg/utils/timeutils.go
```diff
@@ -33,6 +34,14 @@ const (
 	OneDayMs    = 24 * 3600 * 1000
 )
 
+var (
+	// TimeFormats are time formats
+	TimeFormats = []string{
```
gtopper: The comment is a bit redundant I would say 🙂

pkg/utils/timeutils.go
```diff
@@ -68,7 +77,7 @@ func Str2duration(duration string) (int64, error) {
 	return int64(i * multiply), nil
 }
 
-// convert time string to time (unix milisecond)
+// Str2unixTime converts time string to time (unix milisecond)
 func Str2unixTime(timeString string) (int64, error) {
```
gtopper: Also redundant.

pkg/utils/timeutils.go
```diff
@@ -96,11 +105,14 @@ func Str2unixTime(timeString string) (int64, error) {
 		return int64(tint), nil
 	}
 
-	t, err := time.Parse(time.RFC3339, timeString)
-	if err != nil {
-		return 0, errors.Wrap(err, "Not an RFC 3339 time format")
+	for _, format := range TimeFormats {
+		t, err := time.Parse(format, timeString)
```
gtopper: This could potentially slow things, especially in the worst case (data is in the last format).

pkg/utils/timeutils_test.go
tebeka: OK

pkg/utils/timeutils.go
tebeka: Removed

pkg/utils/timeutils.go
tebeka: Comment was there. This is the standard documentation format in Go (start with the function name), see [here](https://blog.golang.org/godoc-documenting-go-code).

pkg/utils/timeutils.go
tebeka: I've benchmarked the new code vs the original version. On my machine it adds 2ns per operation.

Any faster way to do this? In other packages (e.g pandas) there's an optional `format` parameter to the parser function. However didn't want to introduce to many changes.

pkg/utils/timeutils.go
gtopper: Oh well.

pkg/utils/timeutils.go
gtopper: Did you benchmark the worst case (RFC3339Nano input)? What if more formats are added to the list? (what of the worst case then?)

gtopper: Since I don't know the motivation for supporting this change, it's hard to weigh against the potential performance downside.

It does look like this code is only used on the hot path of the Nuclio examples and benchmark, not core production code.

tebeka: > Did you benchmark the worst case (RFC3339Nano input)?
No. I did a `RFC3339` format string since this is the only format the current system uses. If I test the old with `RFC3999` and the new with `RFC3999Nano` then it's about 40ns more.

> What if more formats are added to the list? (what of the worst case then?)
Still about 40ns.  Attaching the benchmark code.

tebeka: > Since I don't know the motivation for supporting this change, it's hard to weigh against the potential performance downside.

Working with @yaronha on https://github.com/v3io/frames, we'd like more than a second granularity on times in the database.

gtopper: Ok, by my own measurements, on my laptop, using a nano timestamp, it takes 351 ns to parse RFC3339, and 390 ns for RFC3339Nano. So 741 ns combined. I guess that's acceptable if you guys actually need it (and it's not e.g. for testing purposes, or something like that).

pkg/utils/timeutils_test.go
```diff
@@ -106,6 +106,17 @@ func (suite *testTimeSuite) TestStr2unixTimeWithNowPlusMinus() {
 	suite.Require().Equal(expectedDuration, endTime-startTime)
 }
 
+func (suite *testTimeSuite) TestStr2unixTimeFRC() {
+	// RFC3339
+	t, err := Str2unixTime("2018-10-04T15:09:36+03:00")
+	suite.Require().Nil(err)
+	suite.Require().True(t != 0)
```
gtopper: Why not check for the actually result using `Equal`?

pkg/utils/timeutils.go
```diff
@@ -33,6 +34,13 @@ const (
 	OneDayMs    = 24 * 3600 * 1000
 )
 
+var (
```
gtopper: No need to use this syntax when only defining a single var.

gshatz:
According to @yaronha we can close this for now

PR #166, opened 2018-10-09
change CLI default log level to info

gshatz:
1. chnage default log level of the CLI to info
2. change the print of omitting sample OOO as part of ingestion from debug to info

### DIFF
--- a/pkg/tsdbctl/add.go
+++ b/pkg/tsdbctl/add.go
@@ -251,9 +251,6 @@ func (ac *addCommandeer) appendMetric(
 
 	ac.rootCommandeer.logger.DebugWith("Adding a sample value to a metric.", "lset", lset, "t", tarray, "v", varray)
 
-	if print {
-		fmt.Println("add:", lset, tarray, varray)
-	}
 	ref, err := append.Add(lset, tarray[0], varray[0])
 	if err != nil {
 		return 0, errors.Wrap(err, "Failed to add a sample value to a metric.")
--- a/pkg/tsdbctl/add.go
+++ b/pkg/tsdbctl/add.go
@@ -159,7 +159,7 @@ func (ac *addCommandeer) add() error {
 			return err
 		}
 
-		_, err = ac.appendMetric(appender, lset, tarray, varray, true)
+		_, err = ac.appendMetric(appender, lset, tarray, varray)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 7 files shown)


pkg/tsdbctl/add.go
```diff
@@ -251,9 +251,6 @@ func (ac *addCommandeer) appendMetric(
 
 	ac.rootCommandeer.logger.DebugWith("Adding a sample value to a metric.", "lset", lset, "t", tarray, "v", varray)
 
-	if print {
```
imakhlin: If you decided to drop this print then go ahead and drop the boolean `print` parameter from the `appendMetric` method.
I believe that this "debug" print was added by @yaronha to reflect the input stream content when using CLI with `stdin` or `file` input. Please consider adding useful log message if necessary (using logger API) to let user know what is going on under the hood (@debug level)

imakhlin:
If you want to drop the `print` then drop it all including the boolean parameter to `appendMetric` method.
Consider using a logger to let user know what is going on under the hood when `stdin` or `file` stream is used as an input.

pkg/tsdbctl/tsdbctl.go
```diff
@@ -191,6 +191,8 @@ func (rc *RootCommandeer) populateConfig(cfg *config.V3ioConfig) error {
 	}
 	if rc.logLevel != "" {
 		cfg.LogLevel = rc.logLevel
+	} else {
+		cfg.LogLevel = "info"
```
Sharon-iguazio: Changing the default logging level from `debug` to `info` should include updating the `DefaultLoggingLevel` const in *pkg/config/config.go* and using this const in relevant locations such as in this `*tsdbctl.go* line. (We already use this in this file, *tsdbctl.go*, for the CLI help text of the global `-v|--log-level` flag.)

imakhlin: Not exactly. We use to set log level to `debug` by default for `-v` option without args in CLI.
If used didn't use `-v` the log level should be `info` (according to @gshatz )

Sharon-iguazio:
Q: I see that we have a default log level of "warn" relating to Nuclio (lines 44–45 in **pkg/utils/container.go**) — 
```
	default:
		logLevel = nucliozap.WarnLevel
```
Don't we want to also change this to info?

pkg/tsdbctl/tsdbctl.go
gshatz: the default of "info" is only for the CLI - any other apps using TSDB should maintain their defaults or get the nuclio-logger default of "warn"

Sharon-iguazio:
Please also fix this typo that I just noticed in the description of the `create` command's `-r|--ingestion-rate` flag: `expecetd` > `expected`. Thanks.

Sharon-iguazio:
(Minor) In the code comment for the `AggregateSeries.colName` struct field in **pkg/aggregate/iterator.go** (line 31), "`cloumn`" should be "`column`" (typo) and "`timeseries`" should be "`time series`" (two words).

PR #167, opened 2018-10-09
Performance test response with latency

imakhlin:
* Add `UpdateMetricLatencyHistogram`
* Add `SendTimeNanoseconds` to Request API


vendor/github.com/v3io/v3io-go-http/container.go
```diff
@@ -2,6 +2,7 @@ package v3io
 
 import (
 	"sync/atomic"
+	"time"
```
gtopper: Is this vendor change an update from upstream or a custom edit?

pkg/appender/store.go
```diff
@@ -158,22 +160,28 @@ func (cs *chunkStore) processGetResp(mc *MetricsCache, metric *MetricState, resp
...
+					cs.performanceReporter.IncrementCounter("PutNameError", 1)
+
+					mc.logger.ErrorWith("Update-name PutItem failed", "metric", metric.key, "err", err)
+				} else {
+					mc.logger.DebugWith("Update name", "name", metric.name, "key", metric.key, "reqid", request.ID)
+				}
 			}
+		} else {
+			// Count errors
+			cs.performanceReporter.IncrementCounter("UpdateMetricError", 1)
```
gtopper: Even log this?

vendor/github.com/v3io/v3io-go-http/container.go
```diff
@@ -123,11 +124,12 @@ func (c *Container) sendRequest(input interface{},
...
-			container:    c,
-			Input:        input,
-			Context:      context,
-			responseChan: responseChan,
+			ID:                  id,
+			container:           c,
+			Input:               input,
+			Context:             context,
+			responseChan:        responseChan,
+			SendTimeNanoseconds: time.Now().UnixNano(),
```
gtopper: Looks like allocation time, not send time.

pkg/querier/querier.go
```diff
@@ -110,6 +110,7 @@ func (q *V3ioQuerier) selectQry(
 			return
 		}
 
+		// Bug: TSDB-37 -> TODO: Create result set while taking in account the step. If step is greater than partition interval it should return correct number of entries
```
gtopper: It's not really clear why `TSDB-37` is related to steps. The ticket itself only mentions partitions. Also not really related to this PR 🙂

vendor/github.com/v3io/v3io-go-http/container.go
imakhlin: edited manually, but should be same as in vendor package

pkg/appender/store.go
```diff
@@ -158,22 +161,30 @@ func (cs *chunkStore) processGetResp(mc *MetricsCache, metric *MetricState, resp
...
-			path := mc.cfg.TablePath + "/names/" + metric.name
-			putInput := v3io.PutItemInput{Path: path, Attributes: map[string]interface{}{}}
-
-			request, err := mc.container.PutItem(&putInput, metric, mc.nameUpdateChan)
-			if err != nil {
-				// Count errors
-				cs.performanceReporter.IncrementCounter("PutNameError", 1)
+	latencyNano := time.Now().UnixNano() - resp.Request().SendTimeNanoseconds
+	cs.performanceReporter.UpdateHistogram("UpdateMetricLatencyHistogram", latencyNano)
+	methodLogger := mc.logger.GetChild("processGetResp")
```
gtopper: I'm not sure allocating a new logger per call is a good idea. It's also not the practice we have, which is to allocate it with the struct.

pkg/appender/store.go
```diff
@@ -158,22 +161,30 @@ func (cs *chunkStore) processGetResp(mc *MetricsCache, metric *MetricState, resp
...
+			if metric.newName {
+				path := mc.cfg.TablePath + "/names/" + metric.name
+				putInput := v3io.PutItemInput{Path: path, Attributes: map[string]interface{}{}}
+
+				request, err := mc.container.PutItem(&putInput, metric, mc.nameUpdateChan)
+				if err != nil {
+					// Count errors
+					cs.performanceReporter.IncrementCounter("PutNameError", 1)
+
+					methodLogger.ErrorWith("Update-name PutItem failed", "metric", metric.key, "err", err)
```
gtopper: Nice. I actually just realized this never returns an error. But if it does one day we'll be ready 😆

pkg/appender/store.go
```diff
@@ -158,22 +161,30 @@ func (cs *chunkStore) processGetResp(mc *MetricsCache, metric *MetricState, resp
...
+				} else {
+					methodLogger.DebugWith("Update name", "name", metric.name, "key", metric.key, "reqid", request.ID)
+				}
 			}
+		} else {
+			methodLogger.Error("Update metric has failed with error: %v", resp.Error)
+			// Count errors
```
gtopper: I think you don't have to add this comment everywhere. It's pretty clear what it does :)

vendor/github.com/v3io/v3io-go-http/container.go
gtopper: That may be true in your usage of it. It's not true in general. But that's something to take up with upstream I guess.

vendor/github.com/v3io/v3io-go-http/container.go
gtopper: Better check it out instead.

gtopper: (because we don't want manually cherry-picked vendor packages if we can avoid them)

vendor/github.com/v3io/v3io-go-http/container.go
gtopper: Alright, that's true given the context of how it's created (supposing it's only created here).

pkg/appender/store.go
imakhlin: This is our logging infra. I can just reduce amount of debug logs or stop using child loggers at all

pkg/appender/store.go
imakhlin: No comments.... We need to plan for hardening iteration for the entire project.

pkg/appender/store.go
gtopper: Again, the practice is to put the derived logger in the receiver (the struct), not recreating it per method invocation. If you want to specify the method, you can use methods such as `InfoWith` to log and include the method as a variable, e.g. `InfoWith("message here", "method", "processGetResp")`.

If you decide we don't really need these logs, that's ok also.

PR #172, opened 2018-10-11
[DOC] CLI: Edit log-level help text & comments

Sharon-iguazio:
PR #170 take 2

This is the updated help output:
```
-v, --log-level string[="debug"]   Verbose output. Add "=<level>" to set the log level -
                                   debug | info | warn | error. For example: -v=warn.
                                   (default - "debug" when using the flag; "info" otherwise) 
```

@gshatz If you're still not willing to merge all commits, let's talk before you close the PR.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -77,7 +77,7 @@ type V3ioConfig struct {
 	// Disabled = true disables the V3IO TSDB configuration in Prometheus and
 	// enables the internal Prometheus TSDB instead
 	Disabled bool `json:"disabled,omitempty"`
-	// Logging level (for verbose output) - "debug" | "info" | "warn" | "error"
+	// Log level - "debug" | "info" | "warn" | "error"
 	LogLevel string `json:"logLevel,omitempty"`
 	// Number of parallel V3IO worker routines
 	Workers int `json:"workers"`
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -51,6 +51,7 @@ const (
 	DefaultAggregationGranularity = "1h"
 	DefaultLayerRetentionTime     = "1y"
 	DefaultSampleRetentionTime    = 0
+	DefaultLogLevel               = "info"
```
gshatz: if you are adding this variable please use it in the tsdbctl.go (line 195) for the default log level, otherwise there is no point in it

Sharon-iguazio: I already use `DefaultLogLevel` in **tsbdct.go** — on line 64 see (https://github.com/v3io/v3io-tsdb/pull/172/files#diff-e82ef7d65c5957422d23f4a3ea81f257R64 — just scroll to the end):
```
"Verbose output. Add \"=<level>\" to set the log level -\ndebug | info | warn | error. For example: -v=warn.\n(default - \""+config.DefaultVerboseLevel+"\" when using the flag; \""+config.DefaultLogLevel+"\" otherwise)")
```

Sharon-iguazio: See new commit eafd800 — use the `config.DefaultLogLevel` flag to set the default log level in the code (**tsdbctl.go** line 195):
`	cfg.LogLevel = config.DefaultLogLevel`

PR #174, opened 2018-10-12
Fix prometheus not showing all wanted series

talIguaz:
`iterSortMerge` used to return reference to the same object for each series which meant that all "series" pointed to the same data and prometheus reduced them to one series on the graph.
Note: This happened only when using `iterSortMerge` which means only when the query is on multiple partitions

### DIFF
--- a/pkg/querier/multipart.go
+++ b/pkg/querier/multipart.go
@@ -92,7 +92,7 @@ func (im *IterSortMerger) Next() bool {
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


yaronha:
@gshatz @talIguaz @imakhlin not sure if/what it fixes, `im.currSeries[:0]` is a Go trick to zero an array w/o re-allocating mem & avoid GC, what you changed no means it allocates mem and doesn't re-use the buffer (i.e. perf & GC overhead) 

talk to me about it

PR #176, opened 2018-10-14
Fix the prom saw bug

talIguaz:
fix bug when querying multi partitions and getting data labelsets OOO

### DIFF
--- a/pkg/querier/querier.go
+++ b/pkg/querier/querier.go
@@ -111,7 +111,7 @@ func (q *V3ioQuerier) selectQry(
 
 		sets := make([]SeriesSet, len(parts))
 		for i, part := range parts {
-			set, err := q.queryNumericPartition(part, name, functions, step, windows, filter)
+			set, err = q.queryNumericPartition(part, name, functions, step, windows, filter)
 			if err != nil {
 				set = nullSeriesSet{}
 				return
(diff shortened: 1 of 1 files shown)


pkg/querier/querier.go
```diff
@@ -119,17 +119,17 @@ func (q *V3ioQuerier) selectQry(
...
-			for i := 0; i < len(sets); i++ {
-				// TODO make it a Go routine per part
-				sorter, err := NewSetSorter(sets[i])
-				if err != nil {
-					set = nullSeriesSet{}
-					return
-				}
-				sets[i] = sorter
+		// Sort each partition
+		// todo: Removed condition that applies sorting only on non range scan queries, to fix bug with serieses coming OOO when querying multi partitions.
```
imakhlin: `todo` -> `TODO` (could be /*single comment*/)

pkg/querier/querier.go
```diff
@@ -119,17 +119,17 @@ func (q *V3ioQuerier) selectQry(
...
+		// todo: Think of a better solution.
+		for i := 0; i < len(sets); i++ {
+			// TODO make it a Go routine per part
+			sorter, err := NewSetSorter(sets[i])
+			if err != nil {
+				set = nullSeriesSet{}
```
imakhlin: log the errors
Consider adding local logger as `methodLogger := q.logger.GetChild("selectQry")` and then using it within the method for logging necessary stuff. Here for example you can do: `methodLogger.Error("Error: %v", err)`

talIguaz: Note that this method has named return types. so the error is returned and it's up to the caller to manage the error. it gets logged there

imakhlin: Just make sure that we don't miss those errors in the log when running from Prom UI or REST

PR #180, opened 2018-10-14
Bug TSDB-43: avg_over_time in prom console returns NaN when the series is not present in all the time range

talIguaz:


pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -471,6 +471,80 @@ func testQueryDataOverlappingWindowCase(test *testing.T, v3ioConfig *config.V3io
...
 
+// Calling Seek instead of next for the first time while iterating over data (TSDB-43)
+func TestIgnoreNaNWhenSeekingAggSeries(t *testing.T) {
+	v3ioConfig, err := tsdbtest.LoadV3ioConfig()
+	if err != nil {
+		t.Fatalf("unable to load configuration. Error: %v", err)
+	}
+	metricsName := "cpu"
+	userLabels := utils.LabelsFromStrings("os", "linux", "iguaz", "yesplease")
+	data := []tsdbtest.DataPoint{{Time: 1532940510000, Value: 300.3},
```
imakhlin: const refTime = 1532940510000

PR #182, opened 2018-10-15
TSDB-46: Prometheus MIN_OVER_TIME function omits the first value in a series

imakhlin:
- Fix iterator
- Improve logging


imakhlin:
Fixes issue TSDB-46

pkg/querier/series.go
```diff
@@ -147,9 +147,19 @@ func (it *v3ioSeriesIterator) Seek(t int64) bool {
 		t = it.mint
 	}
 
+	// Check the first element
+	t0, _ := it.iter.At()
+	if t0 > it.maxt {
+		return false
+	}
+	if t <= t0 {
+		// The cursor (t0) is either on t or just passed t
```
gtopper: Seems like it could also be way passed it :)

pkg/querier/series.go
```diff
@@ -147,9 +147,19 @@ func (it *v3ioSeriesIterator) Seek(t int64) bool {
...
+		return true
+	}
+
 	for {
 		if it.iter.Next() {
-			t0, _ := it.At()
+			t0, _ := it.iter.At()
```
gtopper: So this one line is the actual fix?

promtsdb/promtsdb.go
```diff
@@ -71,26 +85,32 @@ func (q *V3ioPromQuerier) Select(params *storage.SelectParams, oms ...*labels.Ma
...
 // LabelValues returns all potential values for a label name.
-func (q *V3ioPromQuerier) LabelValues(name string) ([]string, error) {
-	return q.q.LabelValues(name)
+func (promQuery *V3ioPromQuerier) LabelValues(name string) ([]string, error) {
+	return promQuery.v3ioQuerier.LabelValues(name)
 }
 
 // Close releases the resources of the Querier.
-func (q *V3ioPromQuerier) Close() error {
+func (promQuery *V3ioPromQuerier) Close() error {
```
gtopper: It's actually somewhat of a convention to abbreviate the receiver :)

promtsdb/promtsdb.go
```diff
@@ -71,26 +85,32 @@ func (q *V3ioPromQuerier) Select(params *storage.SelectParams, oms ...*labels.Ma
...
-func match2filter(oms []*labels.Matcher) (string, string, string) {
-	filter := []string{}
+func match2filter(oms []*labels.Matcher, logger logger.Logger) (string, string, string) {
+	var filter []string
 	agg := ""
 	name := ""
 
+	if logger == nil {
+		// TODO: handle and return the error
+		logger, _ = utils.NewLogger(config.DefaultLogLevel)
```
gtopper: It seems to me that instead of handling the error you could just not instantiate a logger here... I couldn't find any callers that don't pass a logger into this method.

promtsdb/promtsdb.go
```diff
@@ -71,26 +85,32 @@ func (q *V3ioPromQuerier) Select(params *storage.SelectParams, oms ...*labels.Ma
...
+	}
+
 	for _, matcher := range oms {
+		logger.Debug("Matcher: %+v", matcher)
```
gtopper: Seems like it would be more useful to print if it actually matched something.

pkg/querier/querier.go
```diff
@@ -86,16 +87,18 @@ func (q *V3ioQuerier) SelectOverlap(name, functions string, step int64, windows
...
-
 	if err != nil {
 		return nullSeriesSet{}, errors.Wrap(err, "Failed to read/update the TSDB schema.")
 	}
 
+	set = nullSeriesSet{}
+
+	q.logger.Debug("Select query:\n\tMetric: %s\n\tStart Time: %s (%d)\n\tEnd Time: %s (%d)\n\tFunction: %s\n\t"+
+		"Step: %d\n\tFilter: %s\n\tWindows: %v\n\tDisable All Aggr: %t\n\tDisable Client Aggr: %t",
+		name, time.Unix(q.mint/1000, 0).String(), q.mint, time.Unix(q.maxt/1000, 0).String(), q.maxt, functions, step,
```
gtopper: FYI, this will lose sub-second granularity. It's just a log, but I tend to think that's not what you meant to happen.

pkg/querier/series.go
imakhlin: no, here I just call it directly instead of function call

promtsdb/promtsdb.go
imakhlin: it was confusing because of another `q` but I can rename it to `vpq` if you wish

promtsdb/promtsdb.go
imakhlin: correct, I use to pass it everywhere

pkg/querier/querier.go
imakhlin: It also prints the epoch time for accuracy and ability to reuse it in CLI. Human readable time with second precision is enough.

promtsdb/promtsdb.go
imakhlin: it's not designed for unsupported matchers - it does not return no error

gshatz:
This fix is not only for min_over_time - its basically relevant for all functions & aggregations

PR #190, opened 2018-10-19
TSDB-52: Fix partial partition delete 

gshatz:
1. change the default of ignoreErrors to true
2. delete only partitions that completely (not partialy) in the range for delete

### DIFF
--- a/pkg/partmgr/partmgr.go
+++ b/pkg/partmgr/partmgr.go
@@ -279,7 +279,7 @@ func (p *PartitionManager) updatePartitionsFromSchema(schema *config.Schema) err
 	return nil
 }
 
-//if inclusive is true than partial partitions (not fully in range) will be retireved
+//if inclusive is true than partial partitions (not fully in range) will be retireved as well
 func (p *PartitionManager) PartsForRange(mint, maxt int64, inclusive bool) []*DBPartition {
 	var parts []*DBPartition
 	for _, part := range p.partitions {
--- a/pkg/partmgr/partmgr.go
+++ b/pkg/partmgr/partmgr.go
@@ -279,10 +279,11 @@ func (p *PartitionManager) updatePartitionsFromSchema(schema *config.Schema) err
 	return nil
 }
 
-func (p *PartitionManager) PartsForRange(mint, maxt int64) []*DBPartition {
+//if inclusive is true than partial partitions (not fully in range) will be retireved
+func (p *PartitionManager) PartsForRange(mint, maxt int64, inclusive bool) []*DBPartition {
 	var parts []*DBPartition
 	for _, part := range p.partitions {
-		if part.InRange(mint) || part.InRange(maxt) || (mint < part.GetStartTime() && maxt > part.GetEndTime()) {
+		if (mint < part.GetStartTime() && maxt > part.GetEndTime()) || (inclusive && (part.InRange(mint) || part.InRange(maxt))) {
 			parts = append(parts, part)
 		}
 	}
--- a/pkg/tsdb/v3iotsdb.go
+++ b/pkg/tsdb/v3iotsdb.go
@@ -202,18 +202,15 @@ func (a *V3ioAdapter) DeleteDB(deleteAll bool, ignoreErrors bool, fromTime int64
 
 	partitions := a.partitionMngr.PartsForRange(fromTime, toTime, false)
 	for _, part := range partitions {
-		pStart, pEnd := part.GetPartitionRange()
-		if pStart >= fromTime && pEnd <= toTime {
-			a.logger.Info("Delete partition '%s'.", part.GetTablePath())
-			err := utils.DeleteTable(a.logger, a.container, part.GetTablePath(), "", a.cfg.QryWorkers)
-			if err != nil && !ignoreErrors {
-				return errors.Wrapf(err, "Failed to delete partition '%s'.", part.GetTablePath())
-			}
-			// Delete the Directory object
-			err = a.container.Sync.DeleteObject(&v3io.DeleteObjectInput{Path: part.GetTablePath()})
-			if err != nil && !ignoreErrors {
-				return errors.Wrapf(err, "Failed to delete partition object '%s'.", part.GetTablePath())
-			}
+		a.logger.Info("Deleting partition '%s'.", part.GetTablePath())
+		err := utils.DeleteTable(a.logger, a.container, part.GetTablePath(), "", a.cfg.QryWorkers)
+		if err != nil && !ignoreErrors {
+			return errors.Wrapf(err, "Failed to delete partition '%s'.", part.GetTablePath())
+		}
+		// Delete the Directory object
+		err = a.container.Sync.DeleteObject(&v3io.DeleteObjectInput{Path: part.GetTablePath()})
+		if err != nil && !ignoreErrors {
+			return errors.Wrapf(err, "Failed to delete partition object '%s'.", part.GetTablePath())
 		}
 	}
 	a.partitionMngr.DeletePartitionsFromSchema(partitions)
--- a/pkg/tsdb/v3iotsdb.go
+++ b/pkg/tsdb/v3iotsdb.go
@@ -200,7 +200,7 @@ func (a *V3ioAdapter) DeleteDB(deleteAll bool, ignoreErrors bool, fromTime int64
 		toTime = time.Now().Unix() * 1000
 	}
 
-	partitions := a.partitionMngr.PartsForRange(fromTime, toTime)
+	partitions := a.partitionMngr.PartsForRange(fromTime, toTime, false)
 	for _, part := range partitions {
 		pStart, pEnd := part.GetPartitionRange()
 		if pStart >= fromTime && pEnd <= toTime {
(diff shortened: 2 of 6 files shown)


pkg/tsdbctl/delete.go
```diff
@@ -125,9 +129,9 @@ func (dc *delCommandeer) delete() error {
 
 	err = dc.rootCommandeer.adapter.DeleteDB(dc.deleteAll, dc.ignoreErrors, from, to)
 	if err != nil {
-		return errors.Wrapf(err, "Failed to delete TSDB table '%s' in container '%s'.", dc.rootCommandeer.v3iocfg.TablePath, dc.rootCommandeer.v3iocfg.Container)
+		return errors.Wrapf(err, "Failed to delete%s TSDB table '%s' in container '%s'.", partialMsg, dc.rootCommandeer.v3iocfg.TablePath, dc.rootCommandeer.v3iocfg.Container)
 	}
-	fmt.Printf("Successfully deleted TSDB table '%s' from container '%s'.\n", dc.rootCommandeer.v3iocfg.TablePath, dc.rootCommandeer.v3iocfg.Container)
+	fmt.Printf("Successfully deleted%s TSDB table '%s' from container '%s'.\n", partialMsg, dc.rootCommandeer.v3iocfg.TablePath, dc.rootCommandeer.v3iocfg.Container)
```
gtopper: Bad format string.

pkg/tsdbctl/delete.go
```diff
@@ -111,9 +111,13 @@ func (dc *delCommandeer) delete() error {
...
 	}
 
+	partialMsg := ""
+	if !dc.deleteAll {
+		partialMsg = " part of"
+	}
 	if !dc.force {
 		confirmedByUser, err := getConfirmation(
-			fmt.Sprintf("You are about to delete TSDB table '%s' in container '%s'. Are you sure?", dc.rootCommandeer.v3iocfg.TablePath, dc.rootCommandeer.v3iocfg.Container))
+			fmt.Sprintf("You are about to delete%s TSDB table '%s' in container '%s'. Are you sure?", partialMsg, dc.rootCommandeer.v3iocfg.TablePath, dc.rootCommandeer.v3iocfg.Container))
```
imakhlin: add space before %s and remove from the message

pkg/tsdb/v3iotsdb.go
```diff
@@ -202,15 +202,18 @@ func (a *V3ioAdapter) DeleteDB(deleteAll bool, ignoreErrors bool, fromTime int64
...
-		a.logger.Info("Delete partition '%s'.", part.GetTablePath())
-		err := utils.DeleteTable(a.logger, a.container, part.GetTablePath(), "", a.cfg.QryWorkers)
-		if err != nil && !ignoreErrors {
-			return errors.Wrapf(err, "Failed to delete partition '%s'.", part.GetTablePath())
-		}
-		// Delete the Directory object
-		err = a.container.Sync.DeleteObject(&v3io.DeleteObjectInput{Path: part.GetTablePath()})
-		if err != nil && !ignoreErrors {
-			return errors.Wrapf(err, "Failed to delete partition object '%s'.", part.GetTablePath())
+		pStart, pEnd := part.GetPartitionRange()
```
imakhlin: IMHO, if given time interval does contains partial partitions they should never be deleted and user should get notified about.
Only partitions within given interval should be deleted

pkg/tsdb/v3iotsdb.go
```diff
@@ -202,15 +202,18 @@ func (a *V3ioAdapter) DeleteDB(deleteAll bool, ignoreErrors bool, fromTime int64
...
+		if pStart >= fromTime && pEnd <= toTime {
```
imakhlin: add `else` case and log message that some partitions were not deleted because they are not fully included within given gange

pkg/tsdbctl/delete.go
gshatz: it won't work. the %s can be empty string as well, so I don't want 2 spaces there

imakhlin: Just update the message and you'll be able to do what you want.
For example:
```
if !dc.deleteAll {
	partialMsg = " part of"
} else {
	partialMsg = "entire"
}
```
Note, in some cases the given time interval could be greater than your actual database range, so saying "partial" in any case is incorrect.

pkg/partmgr/partmgr.go
```diff
@@ -279,10 +279,11 @@ func (p *PartitionManager) updatePartitionsFromSchema(schema *config.Schema) err
 	return nil
 }
 
-func (p *PartitionManager) PartsForRange(mint, maxt int64) []*DBPartition {
+//if inclusive is true than partial partitions (not fully in range) will be retireved as well
+func (p *PartitionManager) PartsForRange(mint, maxt int64, inclusive bool) []*DBPartition {
```
imakhlin: Update the test. Check with `inclusive=false`
See: partmgr_test.go

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -619,6 +619,139 @@ func TestDeleteTSDB(t *testing.T) {
...
+		data         []tsdbtest.DataPoint
+		expected     []tsdbtest.DataPoint
+		ignoreReason string
+		expectFail   bool
+	}{
+		{desc: "Should delete all table",
+			deleteFrom:     0,
+			deleteTo:       1599999999999,
+			ignorErrors:	true,
+			data:     		[]tsdbtest.DataPoint{{Time: 1538580053000, Value: 222.2},
```
imakhlin: It would be nice to make the timestamps human readable (at least by adding comment)
Examples & Alternatives:
1. Absolute time
```
t := time.Now()
// Round the test time down to the closest hour to get predictable and consistent results
currentRoundedTimeNano := time.Date(t.Year(), t.Month(), t.Day(), t.Hour(), 0, 0, 0, t.Location()).UnixNano()
```
2. Use duration API: `now-2d`, etc.

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -619,6 +619,139 @@ func TestDeleteTSDB(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unable to load configuration. Error: %v", err)
+	}
+
+	testCases := []struct {
+		desc         string
+		deleteFrom   int64
+		deleteTo     int64
+		deleteAll	 bool
+		ignorErrors  bool
```
imakhlin: typo: `ignore`

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -619,6 +619,139 @@ func TestDeleteTSDB(t *testing.T) {
...
+	adapter, teardown := tsdbtest.SetUpWithData(test, v3ioConfig, metricsName, data, userLabels)
+	defer teardown()
+
+	if err := adapter.DeleteDB(false, ignoreErrors, deleteFrom, deleteTo); err != nil {
+		test.Fatalf("Failed to delete DB on teardown. reason: %s", err)
+	}
+
+	qry, err := adapter.Querier(nil, 0, time.Now().Unix() * 1000)
+	if err != nil {
+		if expectFail {
```
imakhlin: You should never expect failure while obtaining a Querier in delete TSDB flow. You actually doesn't have any reasonable scenario where failure is expected in your tests, so why do you need it?

pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -619,6 +619,139 @@ func TestDeleteTSDB(t *testing.T) {
...
+	expected []tsdbtest.DataPoint, expectFail bool) {
+
+
+	adapter, teardown := tsdbtest.SetUpWithData(test, v3ioConfig, metricsName, data, userLabels)
+	defer teardown()
+
+	if err := adapter.DeleteDB(false, ignoreErrors, deleteFrom, deleteTo); err != nil {
+		test.Fatalf("Failed to delete DB on teardown. reason: %s", err)
+	}
+
```
imakhlin: Check also the number of partitions remaining after partial delete.
For example:
```
schema := testutils.CreateSchema(test, "")
pm, err := partmgr.NewPartitionMngr(schema, nil, v3ioConfig)

initiaPartitions := pm.PartsForRange(0, time.Now().Unix() * 1000, false)
initialNumberOfPartitions := len(initiaPartitions)
	
partitionsToDelete := pm.PartsForRange(deleteFrom, deleteTo, false)
	
assert.Equal(test, len(partitionsToDelete), numberOfPartitionsInRange)
	
if err := adapter.DeleteDB(false, ignoreErrors, deleteFrom, deleteTo); err != nil {
	test.Fatalf("Failed to delete DB on teardown. reason: %s", err)
}

remainingParts := pm.PartsForRange(0, time.Now().Unix() * 1000, false)
assert.Equal(test, remainingParts,  initialNumberOfPartitions - numberOfPartitionsInRange)
```

pkg/tsdb/v3iotsdb_integration_test.go
gshatz: it should be absolut, not relative (like now-...)

PR #192, opened 2018-10-21
Support checking objects' content by metric name and labels.

gtopper:
In addition to doing so by object path.

### DIFF
--- a/pkg/tsdbctl/check.go
+++ b/pkg/tsdbctl/check.go
@@ -21,6 +21,7 @@ such restriction.
 package tsdbctl
 
 import (
+	"encoding/json"
 	"fmt"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -28,6 +29,10 @@ import (
 	"github.com/v3io/v3io-tsdb/pkg/chunkenc"
 	"github.com/v3io/v3io-tsdb/pkg/config"
 	"github.com/v3io/v3io-tsdb/pkg/utils"
+	"math"
+	"path"
+	"strconv"
+	"strings"
 	"time"
 )
 
@@ -35,7 +40,12 @@ type checkCommandeer struct {
 	cmd            *cobra.Command
 	rootCommandeer *RootCommandeer
 	objPath        string
+	metricName     string
+	labels         utils.Labels
 	attrs          []string
+	from           string
+	to             string
+	last           string
 }
 
 func newCheckCommandeer(rootCommandeer *RootCommandeer) *checkCommandeer {
@@ -56,11 +66,32 @@ Arguments:
 - <item-path> (string) [Required] Path to a metric item within the TSDB table.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 
-			if len(args) == 0 {
-				return errors.New("The check command requires an item path.")
+			numArgs := len(args)
+
+			if numArgs == 0 {
+				return errors.New("The check command requires a metric name.")
 			}
 
-			commandeer.objPath = args[0]
+			if strings.Contains(args[0], "/") {
+				if len(commandeer.attrs) == 0 {
+					return errors.New("-a (--attrs) is required when specifying a path")
+				}
+				commandeer.objPath = args[0]
+				return commandeer.check()
+			}
+
+			commandeer.metricName = args[0]
+			commandeer.labels = append(commandeer.labels, utils.Label{Name: "__name__", Value: commandeer.metricName})
+
+			if numArgs > 1 {
+				for _, labelStr := range strings.Split(args[1], ",") {
+					splitLabelParts := strings.SplitN(labelStr, "=", 2)
+					if len(splitLabelParts) < 2 {
+						return errors.Errorf("failed to parse '%s' as a label", labelStr)
+					}
+					commandeer.labels = append(commandeer.labels, utils.Label{Name: splitLabelParts[0], Value: splitLabelParts[1]})
+				}
+			}
 
 			// Initialize parameters
 			return commandeer.check()
@@ -68,8 +99,13 @@ Arguments:
 	}
 
 	cmd.Flags().StringSliceVarP(&commandeer.attrs, "attrs", "a", []string{},
-		"[Required] An array of metric-item blob attribute names, as a\ncomma-separated list. For example: \"_v35\"; \"_v22,_v23\".")
-	cmd.MarkFlagRequired("attrs")
+		"An array of metric-item blob attribute names, as a\ncomma-separated list. For example: \"_v35\"; \"_v22,_v23\".")
+	cmd.Flags().StringVarP(&commandeer.to, "end", "e", "",
+		"End (maximum) time for the query, as a string containing an\nRFC3339 time string, a Unix timestamp in milliseconds, or\na relative time of the format \"now\" or \"now-[0-9]+[mhd]\"\n(where 'm' = minutes, 'h' = hours, and 'd' = days).\nExamples: \"2018-09-26T14:10:20Z\"; \"1537971006000\";\n\"now-3h\"; \"now-7d\". (default \"now\")")
+	cmd.Flags().StringVarP(&commandeer.from, "begin", "b", "",
+		"Start (minimum) time for the query, as a string containing\nan RFC3339 time, a Unix timestamp in milliseconds, a\nrelative time of the format \"now\" or \"now-[0-9]+[mhd]\"\n(where 'm' = minutes, 'h' = hours, and 'd' = days), or 0\nfor the earliest time. Examples: \"2016-01-02T15:34:26Z\";\n\"1451748866\"; \"now-90m\"; \"0\". (default = <end time> - 1h)")
+	cmd.Flags().StringVarP(&commandeer.last, "last", "l", "",
+		"Return data for the specified time period before the\ncurrent time, of the format \"[0-9]+[mhd]\" (where\n'm' = minutes, 'h' = hours, and 'd' = days>). When setting\nthis flag, don't set the -b|--begin or -e|--end flags.\nExamples: \"1h\"; \"15m\"; \"30d\" to return data for the last\n1 hour, 15 minutes, or 30 days.")
 
 	commandeer.cmd = cmd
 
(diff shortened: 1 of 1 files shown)


pkg/tsdbctl/check.go
```diff
@@ -56,20 +66,46 @@ Arguments:
...
 			return commandeer.check()
 		},
 	}
 
 	cmd.Flags().StringSliceVarP(&commandeer.attrs, "attrs", "a", []string{},
-		"[Required] An array of metric-item blob attribute names, as a\ncomma-separated list. For example: \"_v35\"; \"_v22,_v23\".")
-	cmd.MarkFlagRequired("attrs")
+		"An array of metric-item blob attribute names, as a\ncomma-separated list. For example: \"_v35\"; \"_v22,_v23\".")
+	cmd.Flags().StringVarP(&commandeer.to, "end", "e", "",
+		"End (maximum) time for the query, as a string containing an\nRFC3339 time string, a Unix timestamp in milliseconds, or\na relative time of the format \"now\" or \"now-[0-9]+[mhd]\"\n(where 'm' = minutes, 'h' = hours, and 'd' = days).\nExamples: \"2018-09-26T14:10:20Z\"; \"1537971006000\";\n\"now-3h\"; \"now-7d\". (default \"now\")")
```
imakhlin: `RFC3339` -> `RFC 3339`

imakhlin: default value (`now`) should be set for the `end` flag

pkg/tsdbctl/check.go
```diff
@@ -56,20 +66,46 @@ Arguments:
...
+	cmd.Flags().StringVarP(&commandeer.from, "begin", "b", "",
+		"Start (minimum) time for the query, as a string containing\nan RFC3339 time, a Unix timestamp in milliseconds, a\nrelative time of the format \"now\" or \"now-[0-9]+[mhd]\"\n(where 'm' = minutes, 'h' = hours, and 'd' = days), or 0\nfor the earliest time. Examples: \"2016-01-02T15:34:26Z\";\n\"1451748866\"; \"now-90m\"; \"0\". (default = <end time> - 1h)")
```
imakhlin: `RFC3339` -> `RFC 3339`
default value (`now-1`) could be set for the `begin` flag

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
-	objPath := fmt.Sprintf("%s/%s", tablePath, cc.objPath)
+
+	var respChans []chan *v3io.Response
+	if cc.metricName == "" {
+		respChan, err := cc.checkByPath(container, tablePath)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check using object path '%s'", cc.objPath)
+		}
+		respChans = []chan *v3io.Response{respChan}
+		objPath := "/" + path.Join("", tablePath, cc.objPath)
```
imakhlin: change to `objPath := path.Join("/", tablePath, cc.objPath)` otherwise you may get "//" at the beginning.
Note `path.Join` already takes care of separators

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
+		}
+	} else {
+		respChans, err = cc.checkByName(container, tablePath)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check using metric name (%s) and labels (%v)", cc.metricName, cc.labels)
+		}
+	}
+
+	for _, respChan := range respChans {
+		resp := <-respChan
```
imakhlin: I think it could hang on waiting on channel - add timeout

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
+		}
+	}
+
+	return nil
+}
+
+func (cc *checkCommandeer) checkByPath(container *v3io.Container, tablePath string) (chan *v3io.Response, error) {
+	var err error
+	respChan := make(chan *v3io.Response, 1)
+	objPath := "/" + path.Join("", tablePath, cc.objPath)
```
imakhlin: change to `objPath := path.Join("/", tablePath, cc.objPath)` otherwise you may get "//" at the beginning.
Note `path.Join` already takes care of separators

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
+	return respChan, nil
+}
+
+func (cc *checkCommandeer) checkByName(container *v3io.Container, tablePath string) ([]chan *v3io.Response, error) {
+
+	var err error
+
+	schema, err := getSchema(cc.rootCommandeer.v3iocfg, container)
+	if err != nil {
+		return nil, err
```
imakhlin: wrap?

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
+	}
+	numBuckets := schema.TableSchemaInfo.ShardingBucketsCount
+	hash := cc.labels.Hash()
+	var from, to int64
+	if cc.from == "" {
+		from = 0
+	} else {
+		from, err = utils.Str2unixTime(cc.from)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse start time")
```
imakhlin: add `cc.from` to the error message

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
+		}
+	}
+	if cc.to == "" {
+		to = math.MaxInt64
+	} else {
+		to, err = utils.Str2unixTime(cc.to)
+		return nil, errors.Wrap(err, "failed to parse end time")
```
imakhlin: same here

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
+	if cc.from == "" {
+		from = 0
+	} else {
+		from, err = utils.Str2unixTime(cc.from)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse start time")
+		}
+	}
+	if cc.to == "" {
+		to = math.MaxInt64
```
imakhlin: shouldn't it be set to `now`?

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
+
+	schema, err := getSchema(cc.rootCommandeer.v3iocfg, container)
+	if err != nil {
+		return nil, err
+	}
+	numBuckets := schema.TableSchemaInfo.ShardingBucketsCount
+	hash := cc.labels.Hash()
+	var from, to int64
+	if cc.from == "" {
+		from = 0
```
imakhlin: from command description follows that default is `now-1h` or `end-time-1h`

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
+		to = math.MaxInt64
+	} else {
+		to, err = utils.Str2unixTime(cc.to)
+		return nil, errors.Wrap(err, "failed to parse end time")
+	}
+	partitionInterval, _ := utils.Str2duration(schema.PartitionSchemaInfo.PartitionerInterval)
+	var respChans []chan *v3io.Response
+	for _, partition := range schema.Partitions {
+		partitionEndTime := partition.StartTime + partitionInterval
+		if partitionEndTime >= from && partition.StartTime <= to {
```
imakhlin: the condition looks wrong to me. 
IMHO it should be `partition.StartTime >= from &&  partitionEndTime <= to`

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
 				}
-				fmt.Printf("Total size=%d, Count=%d, Compression ratio=%.2f\n",
-					bytesCount, count, compressionRatio)
 			}
 		}
 	}
+	return nil
+}
 
+func (cc *checkCommandeer) printValues(bytes []byte) error {
```
imakhlin: Consider moving useful helper functions to a package so they might be reused by others

pkg/tsdbctl/check.go
gtopper: When `from` or `to` falls in the middle of a partition, we want to include that partition.

pkg/tsdbctl/check.go
gtopper: That's the responsibility of v3io-go-http. I'm actually adding a timeout API to it right now. See https://github.com/v3io/v3io-go-http/pull/10

pkg/tsdbctl/check.go
gtopper: I don't think there's a point adding another layer here, just because I broke out a method for readability.

pkg/tsdbctl/check.go
gtopper: I'm pretty sure that's not actually in the command description. Maybe for another command. We want default to be all time.

pkg/tsdbctl/check.go
gtopper: We want default to be all time.

pkg/tsdbctl/check.go
gtopper: That's not a generic utility, just some specific logic.

pkg/tsdbctl/check.go
imakhlin: This is how we did it almost everywhere...

pkg/tsdbctl/check.go
```diff
@@ -91,58 +126,172 @@ func (cc *checkCommandeer) check() error {
...
 	container, tablePath := cc.rootCommandeer.adapter.GetContainer()
-	objPath := fmt.Sprintf("%s/%s", tablePath, cc.objPath)
+
+	var respChans []chan *v3io.Response
+	if cc.metricName == "" {
+		respChan, err := cc.checkByPath(container, tablePath)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check using object path '%s'", cc.objPath)
+		}
+		respChans = []chan *v3io.Response{respChan}
```
talIguaz: Remove unnecessary getItem

PR #198, opened 2018-10-25
Fix metric cache key.

gtopper:


pkg/appender/appender.go
```diff
@@ -95,6 +95,11 @@ func (m *MetricState) error() error {
 	return m.err
 }
 
+type cacheKey struct {
```
gshatz: I think the perfomance-wise, it better to create a new uint64 hash rather than have this struct be the key to the map. WDYT?

PR #205, opened 2018-10-31
CLI check - support fetching the aggregation arrays

talIguaz:


pkg/tsdbctl/check.go
```diff
@@ -230,16 +230,30 @@ func (cc *checkCommandeer) printResponse(resp *v3io.Response) error {
...
+					}
 				}
 			}
 		}
 	}
 	return nil
 }
 
+func (cc *checkCommandeer) printArrays(bytes []byte) {
+	fmt.Print("[")
```
imakhlin: How about building one string and print it once instead of multiple `fmt.Print` calls?

gtopper: +1

PR #209, opened 2018-11-05
fix server-side 'sqr' aggregate being calculated twice

talIguaz:


pkg/tsdb/v3iotsdb_integration_test.go
```diff
@@ -289,6 +289,22 @@ func TestQueryData(t *testing.T) {
 			expected: map[string][]tsdbtest.DataPoint{
 				"count": {{Time: 1537972158402, Value: 1},
 					{Time: 1537973058402, Value: 2}}}},
+
+		{desc: "Should ingest and query server-side aggregates", metricName: "cpu",
+			labels: utils.LabelsFromStrings("os", "linux", "iguaz", "yesplease"),
+			data: []tsdbtest.DataPoint{{Time: 1532940510, Value: 300.3},
```
imakhlin: IMHO, it's much easier to read human readable time.
Consider initialise the timestamp using human readable syntax. For example:
```
t, _ := time.Parse(time.RFC3339, "2018-11-03T00:00:00Z")
```

PR #210, opened 2018-11-05
TSDB-62: Fix incorrect merge of series

gtopper:
due to removal of metric name from hash function.

Regression integration test included.

### DIFF
--- a/pkg/tsdb/tsdb_integration_test.go
+++ b/pkg/tsdb/tsdb_integration_test.go
@@ -23,7 +23,6 @@ such restriction.
 package tsdb_test
 
 import (
-	"fmt"
 	"github.com/stretchr/testify/suite"
 	"github.com/v3io/v3io-tsdb/pkg/tsdb"
 	"github.com/v3io/v3io-tsdb/pkg/tsdb/tsdbtest"
@@ -87,7 +86,6 @@ func (suite *testTsdbSuite) TestAppend() {
 			samples[tv{t: t, v: v}] = struct{}{}
 		}
 		result[key] = samples
-		fmt.Println(key, result[key])
 	}
 
 	expected := map[string]map[tv]struct{}{
--- a/pkg/tsdb/tsdb_integration_test.go
+++ b/pkg/tsdb/tsdb_integration_test.go
@@ -39,40 +39,43 @@ type testTsdbSuite struct {
 func (suite *testTsdbSuite) TestAppend() {
 
 	v3ioConfig, err := tsdbtest.LoadV3ioConfig()
-	suite.Require().Nil(err)
+	suite.Require().NoError(err)
 
 	defer tsdbtest.SetUp(suite.T(), v3ioConfig)()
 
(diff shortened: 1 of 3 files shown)


pkg/utils/labels.go
```diff
@@ -117,6 +117,22 @@ func (ls *Labels) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// Hash returns a hash value for the label set.
+func (ls Labels) HashWithMetricName() uint64 {
```
imakhlin: try to avoid duplicating code.
you can create `func hash (ignoreFields []string)` and call it from both public methods

imakhlin: FYI: latest Prometheus provide veriety of hashing finctions
https://github.com/v3io/prometheus/blob/v3io-prometheus-v2.4.3/pkg/labels/labels.go#L115

pkg/tsdb/tsdb_integration_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	appender, err := adapter.Appender()
+	suite.Require().Nil(err)
+
+	querier, err := adapter.Querier(nil, 0, math.MaxInt64)
+	suite.Require().Nil(err)
+
+	_, err = appender.Add(
+		utils.Labels{utils.Label{Name: "__name__", Value: "AAPL"}, utils.Label{Name: "market", Value: "usa"}},
+		1541030400000,
```
imakhlin: consider using human readable time instead of Unix time to make reviews easier
for example:
```
t, _ := time.Parse(time.RFC3339, "2018-11-03T00:00:00Z")
	fmt.Printf("Time: %v", t.Unix()*1000)
```

pkg/utils/labels.go
gtopper: I don't see any upsides to code sharing in this way, especially with functionality that no one will ever use for any other purpose (ignoreFields).

gtopper: What about the Prometheus code, are you suggesting I use it...?

pkg/tsdb/tsdb_integration_test.go
gtopper: But why would you want to read it? 🙂 The time isn't really the point of the test.

imakhlin: I know, but all the time I do review others code it makes me crazy. It's unclear from Unix time if the sample in same partition range or far enough from other sample to be in different partition, etc.
In this case I know exactly what're you trying to test and at the moment it's clear to me.

pkg/utils/labels.go
imakhlin: Not in this version. After merging 2.4.3 to the master we might take some good things

pkg/tsdb/tsdb_integration_test.go
gtopper: I suppose I could change it to a string to keep you sane 😋

PR #211, opened 2018-11-05
Querier V2

talIguaz:
* Yaron's initial skeleton
* raw data query
* server-side aggregates
* dataframe column inferring


pkg/aggregate/aggregate.go
```diff
@@ -148,6 +148,15 @@ func AggrsFromString(split []string) (AggrType, error) {
 	return aggrList, nil
 }
 
+func AggregateFromString(aggrString string) (AggrType, error) {
+	trimmed := strings.TrimSpace(aggrString)
+	aggr, ok := aggrTypeString[trimmed]
```
imakhlin: - [ ] make it reusable (see NewAggregationParams, etc.)

pkg/aggregate/aggregate.go
```diff
@@ -223,3 +232,140 @@ func (a AggregatesList) Clear() {
 		aggr.Clear()
 	}
 }
+
+func GetHiddenAggregates(mask AggrType, requestedAggregates []AggrType) []AggrType {
```
imakhlin: consider replacing `[]AggrType` with single aggregated `AggrType` (bitwise OR)
It will make code simpler and reduce complexity of checks

pkg/aggregate/aggregate.go
```diff
@@ -223,3 +232,140 @@ func (a AggregatesList) Clear() {
...
+	}
+	return hiddenAggregates
+}
+
+func GetHiddenAggregatesWithCount(mask AggrType, requestedAggregates []AggrType) []AggrType {
+	mask |= aggrTypeCount
+	return GetHiddenAggregates(mask, requestedAggregates)
+}
+
+func ContainsAggregate(list []AggrType, item AggrType) bool {
```
imakhlin: see my previous comment, you can make it much more efficient with bitwise operations

pkg/aggregate/aggregationParams.go
```diff
@@ -0,0 +1,88 @@
...
+func (as *AggregationParams) GetFunctions() []AggrType {
+	return as.functions
+}
+
+func (as *AggregationParams) NumFunctions() int {
+	return len(as.functions)
+}
+
+func (as *AggregationParams) toAttrName(aggr AggrType) string {
+	return "_" + as.colName + "_" + aggr.String()
```
imakhlin: use `fmt.Sprintf(...)`

pkg/aggregate/iterator.go
```diff
@@ -41,6 +41,28 @@ type AggregateSeries struct {
 	overlapWindows []int      // a list of overlapping windows (* interval), e.g. last 1hr, 6hr, 12hr, 24hr
 }
 
+func StrToAggr(functions string) (AggrType, []AggrType, error) {
```
imakhlin: I feel `de ja vu` ...

pkg/formatter/formatters.go
```diff
@@ -121,3 +121,29 @@ func (f simpleJsonFormatter) Write(out io.Writer, set querier.SeriesSet) error {
...
+	baseFormatter
+}
+
+func (f nilFormatter) Write(out io.Writer, set querier.SeriesSet) error {
+	var count int
+	for set.Next() {
+		count++
+		series := set.At()
+		iter := series.Iterator()
+		for iter.Next() {
```
imakhlin: what's the purpose of this empty loop?

pkg/formatter/formatters.go
```diff
@@ -121,3 +121,29 @@ func (f simpleJsonFormatter) Write(out io.Writer, set querier.SeriesSet) error {
...
+		if iter.Err() != nil {
+			return iter.Err()
+		}
+	}
+
+	if set.Err() != nil {
+		return set.Err()
+	}
+
+	fmt.Fprintf(out, "got %v unique label sets\n", count)
```
imakhlin: log it if you need it

pkg/pquerier/Series.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+func NewDataFrameColumnSeries(indexColumn, dataColumn, countColumn Column, labels utils.Labels, hash uint64) *DataFrameColumnSeries {
+	s := &DataFrameColumnSeries{dataColumn: dataColumn, indexColumn: indexColumn, CountColumn: countColumn, labels: labels, key: hash}
+	s.iter = &DataFrameColumnSeriesIterator{indexColumn: indexColumn, dataColumn: dataColumn, countColumn: countColumn, currentIndex: -1}
+	return s
+}
+
+// This series converts two columns into a series of time-value pairs
+type DataFrameColumnSeries struct {
```
imakhlin: you can just have the `DataFrameColumnSeries` as a member instead of copying all it properties

pkg/pquerier/Series.go
```diff
@@ -0,0 +1,103 @@
...
+	dataColumn  Column
+	indexColumn Column
+	CountColumn Column // Count Column is needed to filter out empty buckets
+	labels      utils.Labels
+	key         uint64
+	iter        SeriesIterator
+}
+
+func (s *DataFrameColumnSeries) Labels() utils.Labels {
+	s.labels = append(s.labels, utils.LabelsFromStrings(aggregate.AggregateLabel, s.dataColumn.GetColumnSpec().function.String())...)
```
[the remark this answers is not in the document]
talIguaz: take a look at aggregate.go:112, we override the `String` method of aggrType

pkg/pquerier/Series.go
```diff
@@ -0,0 +1,103 @@
...
+	it.currentIndex = it.getNextValidCell(it.currentIndex)
+
+	// It is enough to only check one of the columns since we assume they are both the same size
+	return it.currentIndex < it.indexColumn.Len()
+}
+
+func (it *DataFrameColumnSeriesIterator) Err() error { return it.err }
+
+func (it *DataFrameColumnSeriesIterator) getNextValidCell(from int) (nextIndex int) {
+	for nextIndex = from + 1; nextIndex < it.dataColumn.Len() && !it.doesCellHasData(nextIndex); nextIndex++ {
```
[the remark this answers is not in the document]
talIguaz: You should initialize this using `NewDataFrameColumnSeries` which in that case we automatically set the initial index to `-1`. The thing is all this iterators are designed so that you call `Next` before the first `At`, that's why the initial index is -1 and not 0

pkg/pquerier/Series.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (it *DataFrameColumnSeriesIterator) doesCellHasData(cell int) bool {
+	// In case we don't have a count column (for example while down sampling) check if there is a real value at `cell`
+	if it.countColumn == nil {
+		f, err := it.dataColumn.FloatAt(cell)
+		if err != nil {
+			it.err = err
+			return false
+		}
+		return !math.IsNaN(f)
```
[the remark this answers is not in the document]
talIguaz: There should not be `Inf` here. the way the columns are built (take a look at the collector logic) every cell has either a value or NaN.

pkg/pquerier/chunkIterator.go
```diff
@@ -0,0 +1,240 @@
...
+	for _, attr := range item.query.attrs {
+
+		// In case we get both raw chunks and server aggregates, only go over the chunks.
+		if !strings.Contains(attr, aggregateAttributePrefix) {
+			values := item.fields[attr]
+			if values != nil {
+				bytes := values.([]byte)
+				chunk, err := chunkenc.FromData(it.log, chunkenc.EncXOR, bytes, 0)
+				if err != nil {
+					it.log.ErrorWith("Error reading chunk buffer", "columns", item.query.attrs, "err", err)
```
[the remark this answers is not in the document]
talIguaz: it was pretty much copied from the old querier. currently "faulted" chunks will be ignored and not showed in the result. not sure if we really want to fail the whole query.

pkg/formatter/formatters.go
talIguaz: wanted to exhaust the iterator, but isnt really needed ill delete it.

pkg/pquerier/Series.go
talIguaz: It basically exposes an iterator so once i iterate it once it's useless. so if i'll call it twice second time is empty result.
Plus, there is a new `DataFrameColumnSeries` per every data column and i only pass pointers to the relevant columns.

pkg/pquerier/chunkIterator.go
```diff
@@ -0,0 +1,240 @@
...
+	log        logger.Logger
+
+	prevT int64
+	prevV float64
+}
+
+func newRawChunkIterator(item qryResults, log logger.Logger) SeriesIterator {
+	maxt := item.query.maxt
+	maxTime := item.fields["_maxtime"]
+	if maxTime != nil && int64(maxTime.(int)) < maxt {
```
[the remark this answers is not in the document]
talIguaz: no can do. the value behind the `maxTime interface{}` is of type `int` therefor can not be casted directly to int64. (getting conversion error and panics)

pkg/aggregate/aggregationParams.go
```diff
@@ -0,0 +1,88 @@
...
+		Interval:       interval,
+		overlapWindows: windows,
+	}
+
+	return &newAggregateSeries, nil
+}
+
+func (as *AggregationParams) CanAggregate(partitionAggr AggrType) bool {
+	// keep only real aggregates
+	aggrMask := 0x7f & as.aggrMask
```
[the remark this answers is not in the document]
talIguaz: talked about this in person

pkg/pquerier/Series.go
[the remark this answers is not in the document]
imakhlin:
```
func (it *dataFrameColumnSeriesIterator) Next() bool {
	if it.err != nil {
		return false
	}
	
	var hasMore = false
	it.currentIndex, hasMore = it.getNextValidCell(it.currentIndex)

	// It is enough to only check one of the columns since we assume they are both the same size
	return hasMore
}

func (it *dataFrameColumnSeriesIterator) getNextValidCell(from int) (int, bool) {
	nextIndex := from + 1 // starts with -1
	for ; nextIndex < it.dataColumn.Len() && !it.doesCellHasData(nextIndex); nextIndex++ {
	}

	if nextIndex < it.dataColumn.Len() {
		return nextIndex, true
	} else {
		return from, false
	}
}
```

pkg/pquerier/chunkIterator.go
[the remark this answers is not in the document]
imakhlin: we can make it configurable...
`failOnError=true`

pkg/pquerier/frames.go
```diff
@@ -194,13 +209,49 @@ type dataFrame struct {
 	isRawColumnsGenerated bool
 	rawColumns            []Series
 
-	columns      []Column
-	index        Column
-	columnByName map[string]int // name -> index in columns
+	columnsTemplates []columnMeta
```
imakhlin: `columnsTemplates` -> `columnTemplates`

pkg/pquerier/collector.go
```diff
@@ -91,6 +91,11 @@ func rawCollector(ctx *selectQueryContext, res *qryResults) {
 }
 
 func aggregateClientAggregates(ctx *selectQueryContext, res *qryResults) {
+	err := res.frame.addMetricIfNotExist(res.name, ctx.getResultBucketsSize(), res.IsServerAggregates())
```
imakhlin: Reuse

pkg/pquerier/collector.go
```diff
@@ -0,0 +1,184 @@
...
+				col.SetDataAt(int(currentCell), v)
+			}
+		}
+	}
+}
+
+func aggregateServerAggregates(ctx *selectQueryContext, res *qryResults) {
+	for _, col := range res.frame.columns {
+		if col.GetColumnSpec().metric == res.name &&
+			col.GetColumnSpec().function != 0 &&
```
imakhlin: Add `aggrTypeNone AggrType = 0`

pkg/pquerier/collector.go
```diff
@@ -0,0 +1,184 @@
...
+			array, ok := res.fields[aggregate.ToAttrName(col.GetColumnSpec().function)]
+			if !ok {
+				ctx.logger.Warn("requested function %v was not found in response", col.GetColumnSpec().function)
+			} else {
+				// go over the byte array and convert each uint as we go to save memory allocation
+				bytes := array.([]byte)
+				for i := 16; i+8 <= len(bytes); i += 8 {
+					val := binary.LittleEndian.Uint64(bytes[i : i+8])
+					currentValueIndex := (i - 16) / 8
+					currentValueTime := res.query.partition.GetStartTime() + int64(currentValueIndex+1)*res.query.aggregationParams.GetRollupTime()
```
imakhlin: IMHO, better to move function calls out of the loop. Most likely that compiler will optimize it for you but still see no reason for coding it this way

pkg/pquerier/collector.go
```diff
@@ -0,0 +1,184 @@
...
+						floatVal = math.Float64frombits(val)
+					}
+					col.SetDataAt(int(currentCell), floatVal)
+				}
+			}
+		}
+	}
+}
+
+func downsampleRawData(ctx *selectQueryContext, res *qryResults,
```
imakhlin: You can use the DataPoint struct instead of [`t`,`v`]
It was used for tests only so far, but feel free to move it into the library packages

pkg/pquerier/collector.go
```diff
@@ -0,0 +1,184 @@
...
+	previousPartitionLastTime int64, previousPartitionLastValue float64) (int64, float64) {
+	var lastT int64
+	var lastV float64
+	it := newRawChunkIterator(res, nil).(*rawChunkIterator)
+	col := res.frame.columns[res.frame.columnByName[res.name]]
```
imakhlin: Why not calling `func (d *dataFrame) Column(name string) (Column, error)`?

pkg/pquerier/frames.go
```diff
@@ -0,0 +1,660 @@
...
+func (fi *frameIterator) Err() error {
+	return fi.err
+}
+
+// data frame, holds multiple value columns and an index (time) column
+func NewDataFrame(columnsSpec []columnMeta, indexColumn Column, lset utils.Labels, hash uint64, isRawQuery, getAllMetrics bool, columnSize int, useServerAggregates, showAggregateLabel bool) (*dataFrame, error) {
+	df := &dataFrame{lset: lset, hash: hash, isRawSeries: isRawQuery, showAggregateLabel: showAggregateLabel}
+	// is raw query
+	if isRawQuery {
+		df.columnByName = make(map[string]int, 100)
```
imakhlin: magic number
isn't it a variable?

pkg/pquerier/frames.go
```diff
@@ -0,0 +1,660 @@
...
+// will be converted to:
+// time		metric1		metric2
+//	t0		  v0		  NaN
+//	t1		  NaN		  v2
+//	t2		  v1		  v3
+//
+func (d *dataFrame) rawSeriesToColumns() {
+	var timeData []int64
+	columns := make([][]float64, len(d.rawColumns))
+	areExhausted := make([]bool, len(d.rawColumns))
```
imakhlin: make it simple integer counter and reduce by 1 when done until it equals 0

pkg/pquerier/frames.go
```diff
@@ -0,0 +1,660 @@
...
+
+	for _, rawSeries := range d.rawColumns {
+		rawSeries.Iterator().Next()
+		t, _ := rawSeries.Iterator().At()
+		if t < nextTime {
+			nextTime = t
+		}
+	}
+
+	for !allValuesTrue(areExhausted) {
```
imakhlin: `for remaining > 0 { ...`

pkg/pquerier/frames.go
```diff
@@ -0,0 +1,660 @@
...
+		col.SetData(columns[i], numberOfRows)
+		d.columns[i] = col
+	}
+
+	d.isRawColumnsGenerated = true
+}
+
+func (d *dataFrame) shouldGenerateRawColumns() bool { return d.isRawSeries && !d.isRawColumnsGenerated }
+
+func allValuesTrue(indicators []bool) bool {
```
imakhlin: inefficient and unnecessary

pkg/pquerier/frames.go
```diff
@@ -0,0 +1,660 @@
...
+	if !dc.isValidIndex(i) {
+		return fmt.Errorf("index %d out of bounds [0:%d]", i, dc.size)
+	}
+
+	switch reflect.TypeOf(dc.data) {
+	case IntType:
+		dc.data.([]int64)[i] = value.(int64)
+	case FloatType:
+		// Update requested cell, only if not trying to override an existing value with NaN
+		if !(math.IsNaN(value.(float64)) && dc.data.([]float64)[i] != 0) {
```
imakhlin: Pay attention that `0` is a valid value. Maybe you need to initialize the data arrays with `NaN`

pkg/pquerier/frames.go
```diff
@@ -0,0 +1,660 @@
...
+}
+func (c *ConcreteColumn) SetDataAt(i int, val interface{}) error {
+	if !c.isValidIndex(i) {
+		return fmt.Errorf("index %d out of bounds [0:%d]", i, c.size)
+	}
+
+	c.data[i] = c.setFunc(c.data[i], val)
+	return nil
+}
+func (c *ConcreteColumn) GetInterpolationFunction() (InterpolationFunction, int64) { return nil, 0 }
```
imakhlin: consider implementing NOP function or return an error in case you don't want it to be used

pkg/pquerier/frames.go
```diff
@@ -0,0 +1,660 @@
...
+		return 0, fmt.Errorf("index %d out of bounds [0:%d]", i, c.size)
+	}
+
+	value, err := c.function(c.dependantColumns, i)
+	if err != nil {
+		return 0, err
+	}
+	return value.(int64), nil
+}
+func (c *virtualColumn) GetInterpolationFunction() (InterpolationFunction, int64) { return nil, 0 }
```
imakhlin: same here

pkg/pquerier/interpolate.go
```diff
@@ -0,0 +1,101 @@
...
+	return tseek, vprev
+}
+
+// return the same value as in the next point
+func projectNext(tprev, tnext, tseek int64, vprev, vnext float64) (int64, float64) {
+	return tseek, vnext
+}
+
+// linear estimator (smooth graph)
+func projectLin(tprev, tnext, tseek int64, vprev, vnext float64) (int64, float64) {
```
imakhlin: `Lin` -> `Linear`

pkg/pquerier/querier.go
```diff
@@ -0,0 +1,245 @@
...
+
+	disableAllAggr    bool
+	disableClientAggr bool
+}
+
+func (s *SelectParams) getRequestedColumns() []RequestedColumn {
+	if s.RequestedColumns != nil {
+		return s.RequestedColumns
+	}
+	functions := strings.Split(s.Functions, ",")
```
imakhlin: trim?

pkg/pquerier/querier.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+
+		iter, err = selectContext.start(parts, params)
+		return
+	})
+
+	return
+}
+
+func IsPowerOfTwo(x int) bool {
```
imakhlin: Scope: private

pkg/pquerier/querier.go
```diff
@@ -0,0 +1,245 @@
...
+		} else {
+			result, err = q.getLabelValues(labelKey)
+		}
+	})
+	return
+}
+
+func (q *V3ioQuerier) getMetricNames() ([]string, error) {
+	input := v3io.GetItemsInput{
+		Path:           q.cfg.TablePath + "/names/",
```
imakhlin: image

pkg/pquerier/querier.go
```diff
@@ -0,0 +1,245 @@
...
+		metricNames = append(metricNames, iter.GetField(config.ObjectNameAttrName).(string))
+	}
+
+	sort.Sort(sort.StringSlice(metricNames))
+
+	if iter.Err() != nil {
+		q.logger.InfoWith("failed to read metric names; returning an empty list.", "err", iter.Err().Error())
+	}
+
+	return metricNames, nil
```
imakhlin: return `iter.Err()`

pkg/pquerier/querier.go
```diff
@@ -0,0 +1,245 @@
...
+			// to what we're looking for, save the label value in the map.
+			// Use a map to prevent duplicates.
+			if len(splitLabel) == 2 && splitLabel[0] == labelKey {
+				labelValuesMap[splitLabel[1]] = struct{}{}
+			}
+		}
+	}
+
+	if iter.Err() != nil {
+		q.logger.InfoWith("failed to read label values, returning empty list", "err", iter.Err().Error())
```
imakhlin: same comment

pkg/pquerier/query_integration_test.go
```diff
@@ -0,0 +1,1466 @@
...
+	adapter, err := tsdb.NewV3ioAdapter(suite.v3ioConfig, nil, nil)
+	if err != nil {
+		suite.T().Fatalf("failed to create v3io adapter. reason: %s", err)
+	}
+
+	labels1 := utils.LabelsFromStringList("os", "linux")
+	labels2 := utils.LabelsFromStringList("os", "mac")
+	numberOfEvents := 10
+	eventsInterval := 60 * 1000
+	baseTime := time.Now().UnixNano()/1000000 - int64(numberOfEvents*eventsInterval)
```
imakhlin: Reuse 
```
func nanosToMillis(nanos int64) int64 {
	millis := nanos / int64(time.Millisecond)
	return millis
}
```

pkg/pquerier/select.go
```diff
@@ -0,0 +1,435 @@
...
+
+	if s.mint > mint {
+		mint = s.mint
+	}
+
+	for metric := range s.columnsSpecByMetric {
+		var aggregationParams *aggregate.AggregationParams
+		functions, requestAggregatesAndRaw := s.metricsAggregatesToString(metric)
+
+		//if functions == "" && step > 0 && step >= partition.RollupTime() && partition.AggrType().HasAverage() {
```
imakhlin: remove unused code

pkg/pquerier/select.go
```diff
@@ -0,0 +1,435 @@
...
+		s.requestChannels[i] = newChan
+
+		// Increment the WaitGroup counter.
+		s.wg.Add(1)
+		go func(index int) {
+			mainCollector(s, s.requestChannels[index])
+		}(i)
+	}
+
+	return nil
```
imakhlin: should return errors chanel

pkg/pquerier/select.go
```diff
@@ -0,0 +1,435 @@
...
+		// read label set
+		lsetAttr, lok := query.GetField(config.LabelSetAttrName).(string)
+		if !lok {
+			lsetAttr = "UNKNOWN"
+		}
+		if !lok || !nok {
+			s.logger.Error("Error in initLabels; bad field values.")
+		}
+
+		splitLset := strings.Split(lsetAttr, ",")
```
imakhlin: reuse and trim

pkg/tsdbctl/query.go
```diff
@@ -167,6 +170,59 @@ func (qc *queryCommandeer) query() error {
...
+	_, err = formatter.NewFormatter(qc.output, nil)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to start formatter '%s'.", qc.output)
+	}
+
+	var count int
+
+	for set.Next() {
+		series := set.At()
+		//fmt.Println("labels: ", set.At().Labels())
```
imakhlin: remove unnecessary debug prints

pkg/pquerier/querier.go
talIguaz: In some part we want to automatically include the count aggregate. and in this case we do not want to. that's why it's wasn't reused. i can move the functions parsing to the `aggregate` pkg if it will make you feel better 😅

talIguaz: Also the `RequestedColumn` still stores the function as string, so it's not really being parsed rather than just splitting the aggregates

pkg/pquerier/select.go
talIguaz: the error channel is part of the query context so it's accessible there anyway

pkg/appender/store.go
```diff
@@ -168,7 +168,7 @@ func (cs *chunkStore) processGetResp(mc *MetricsCache, metric *MetricState, resp
 	if resp.Error != nil {
 		if utils.IsNotExistsError(resp.Error) {
 			if metric.newName {
-				path := mc.cfg.TablePath + "/names/" + metric.name
+				path := mc.cfg.TablePath + config.NamesDirectory + metric.name
```
imakhlin: https://play.golang.org/p/yXkMs74DJiT

pkg/config/config.go
```diff
@@ -68,6 +68,8 @@ const (
 	AggregateAttrPrefix = "_v_"
 
 	PrometheusMetricNameAttribute = "__name__"
+
+	NamesDirectory = "/names/"
```
imakhlin: Use path API
https://play.golang.org/p/yXkMs74DJiT

pkg/pquerier/frames.go
```diff
@@ -354,7 +354,7 @@ func (d *dataFrame) rawSeriesToColumns() {
 		}
 	}
 
-	for !allValuesTrue(areExhausted) {
+	for nonExhaustedIterators != 0 {
```
imakhlin: `nonExhaustedIterators > 0`

pkg/pquerier/frames.go
```diff
@@ -483,6 +474,10 @@ func (dc *dataColumn) initializeData(dataType DType) {
 		dc.data = make([]int64, dc.size)
 	case FloatType:
 		dc.data = make([]float64, dc.size)
+		floats := dc.data.([]float64)
```
imakhlin: Use array copy for better performance if this is frequent operation
Look at pkg/aggregate/iterator.go:366 for example

pkg/pquerier/querier.go
```diff
@@ -158,7 +159,7 @@ func (q *V3ioQuerier) LabelValues(labelKey string) (result []string, err error)
 
 func (q *V3ioQuerier) getMetricNames() ([]string, error) {
 	input := v3io.GetItemsInput{
-		Path:           q.cfg.TablePath + "/names/",
+		Path:           q.cfg.TablePath + config.NamesDirectory,
```
imakhlin: use path API

pkg/pquerier/frames.go
talIguaz: It happens once for every metric when query specifically for raw data.
if you select `max(cpu), avg(cpu), sqr(cpu), cpu, avg(diskio), diksio` all in one query it will happen twice

imakhlin:
Please update performance comparison reports with V1 vs V2

PR #212, opened 2018-11-06
fix delete last partition on deleteAll 

talIguaz:


pkg/tsdb/v3iotsdb.go
```diff
@@ -195,7 +195,8 @@ func (a *V3ioAdapter) DeleteDB(deleteAll bool, ignoreErrors bool, fromTime int64
 	if deleteAll {
 		// Ignore time boundaries
 		fromTime = 0
-		toTime = time.Now().Unix() * 1000
+		partitionInterval, _ := utils.Str2duration(a.partitionMngr.GetConfig().PartitionSchemaInfo.PartitionerInterval)
```
imakhlin: handle the error anyway

pkg/tsdb/v3iotsdb.go
```diff
@@ -195,7 +195,8 @@ func (a *V3ioAdapter) DeleteDB(deleteAll bool, ignoreErrors bool, fromTime int64
...
+		toTime = time.Now().Unix()*1000 + partitionInterval
 	}
```
gshatz: add test case to v3iotsdb_integration_test.goTestDeleteTable

pkg/tsdb/v3iotsdb.go
gshatz: instead of adding the last partition just give max_int as the toTime

PR #218, opened 2018-11-07
Use `tsdbtest.TestParams` to control the input params and flow of an integration test

imakhlin:
- Use tsdbtest.LoadV3ioConfig() by default to initialize v3io config for integration tests
- Pass `TestParams` as an argument to the tear down function, so you can change tear down behavior of a test
- Following options supported out of the box, other could be added on demand:
  - [x] OptDropTableOnTearDown
  - [x] OptIgnoreReason
  - [x] OptTimeSeries
  - [x] OptV3ioConfig


gshatz:
LGTM, but I prefer @gtopper or @talIguaz  to review as well

imakhlin:
I need to spend some time to make it nice and useful. Definitely not in this PR.

PR #219, opened 2018-11-08
Add build info and version info to TSDB at build time

imakhlin:
Example:
$ **tsdbctl ver**
```
tsdbctl build information:
  Build time: 2018-11-12T11:48:34+0200
  OS: darwin
  Architecture: amd64
  Version: f970431
  SHA: f970431b3aef562119be3d360f1a8c06e963c60a
  Branch: tsdb.revision
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 # All top-level dirs except for vendor/.
 TOPLEVEL_DIRS=`ls -d ./*/. | grep -v '^./vendor/.$$' | sed 's/\.$$/.../'`
 TOPLEVEL_DIRS_GOFMT_SYNTAX=`ls -d ./*/. | grep -v '^./vendor/.$$'`
+TOPLEVEL_DIRS_IMPI_SYNTAX=`ls -d ./*/. | grep -v '^./vendor/.$$' | sed 's/$$/../'`
 
 GIT_SHA := $(shell git rev-parse HEAD)
 GIT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
@@ -59,7 +60,18 @@ build: get
 .PHONY: lint
 lint:
 ifeq ($(shell gofmt -l $(TOPLEVEL_DIRS_GOFMT_SYNTAX)),)
-	# lint OK
+	# gofmt OK
 else
 	$(error Please run `go fmt ./...` to format the code)
 endif
+	@echo Installing linters...
+	go get -u github.com/pavius/impi/cmd/impi
+
+	@echo Verifying imports...
+	$(GOPATH)/bin/impi \
+		--local github.com/iguazio/provazio \
+		--skip pkg/controller/apis \
+		--skip pkg/controller/client \
+		--scheme stdLocalThirdParty \
+		$(TOPLEVEL_DIRS_IMPI_SYNTAX)
+	# Imports OK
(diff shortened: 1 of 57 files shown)


gshatz:
Like!
is there a way for us to invoke this from nuclio?

pkg/config/config.go
```diff
@@ -56,10 +56,31 @@ const (
 	DefaultVerboseLevel           = "debug"
 )
 
+type BuildInfo struct {
+	BuildTime    string `json:"os,omitempty"`
+	Os           string `json:"os,omitempty"`
```
gtopper: C&P error in json tag.

Makefile
```diff
@@ -14,6 +20,18 @@ GOPATH ?= $(shell go env GOPATH)
...
 
+BUILD_TIME := $(shell date +%FT%T%z)
+CONFIG_PKG=github.com/v3io/v3io-tsdb/pkg/config
+
+BUILD_OPTS := -ldflags " \
+  -X $(CONFIG_PKG).osys=$(GOOS) \
+  -X $(CONFIG_PKG).architecture=$(GOARCH) \
+  -X $(CONFIG_PKG).version=$(GIT_REVISION) \
+  -X $(CONFIG_PKG).revision=$(GIT_SHA_SHORT) \
+  -X $(CONFIG_PKG).branch=$(GIT_BRANCH) \
```
gtopper: Best wrap every value with quotes. Better safe than soggy ☔️

pkg/config/config.go
```diff
@@ -56,10 +56,31 @@ const (
...
 	once     sync.Once
 	failure  error
+
+	BuildMetadta = &BuildInfo{
+		BuildTime:    buildTime,
+		Os:           osys,
+		Architecture: architecture,
+		Version:      version,
+		Revision:     revision,
+		Branch:       branch,
```
gtopper: I know it's a minor thing, but can we have these in the same order everywhere? 😅

pkg/config/config.go
```diff
@@ -111,6 +132,8 @@ type V3ioConfig struct {
 	// Don't aggregate from raw chunks, for use when working as a Prometheus
 	// TSDB library
 	DisableClientAggr bool `json:"disableClientAggr,omitempty"`
+	// Build Info
+	BuildInfo *BuildInfo `json:"buildInfo,omitempty"`
```
gtopper: I think not only can we ditch the pointer, this could be a convenient place to use struct "inheritance" notation.

gtopper: e.g.:
```
type A struct {
  B
  extra uint64
}
```

gtopper: But not all that important.

pkg/tsdbctl/version.go
```diff
@@ -0,0 +1,61 @@
...
+	rootCommandeer *RootCommandeer
+}
+
+func newVersionCommandeer(rc *RootCommandeer) *versionCommandeer {
+	commandeer := &versionCommandeer{
+		rootCommandeer: rc,
+	}
+
+	cmd := &cobra.Command{
+		Aliases: []string{"ver"},
```
gtopper: Very lazy alias 😛

pkg/tsdbctl/version.go
```diff
@@ -0,0 +1,61 @@
...
+		Use:     "version",
+		Hidden:  false,
+		Short:   "Displays version",
+		Long:    "Displays version information",
```
gtopper: I think we don't need a short and a long here if the difference between them is going to be the word "information".

imakhlin:
@Sharon-iguazio Please review the wording

PR #220, opened 2018-11-12
Pre-aggregation / cross-label aggregation ingestion.

gtopper:
Adding this functionality to `create` and `query` will be done separately.

### DIFF
--- a/pkg/utils/labels.go
+++ b/pkg/utils/labels.go
@@ -40,11 +40,6 @@ type Label struct {
 	Name, Value string
 }
 
-type LabelIfc interface {
-	GetName() string
-	GetValue() string
-}
-
 // Labels is a sorted set of labels. Order has to be guaranteed upon
 // instantiation.
 type Labels []Label


pkg/appender/ingest.go
```diff
@@ -279,13 +279,15 @@ func (mc *MetricsCache) handleResponse(metric *MetricState, resp *v3io.Response,
 	} else {
 		// Handle Update Expression responses
 		if resp.Error == nil {
-			// Set fields so next write won't include redundant info (bytes, labels, init_array)
-			metric.store.ProcessWriteResp()
+			if metric.store.chunks[0] != nil {
```
imakhlin: add `aggregatesOnly=true` to the store instead

pkg/appender/store.go
```diff
@@ -131,13 +133,15 @@ func (cs *chunkStore) getChunksState(mc *MetricsCache, metric *MetricState) (boo
 	if err != nil {
 		return false, err
 	}
-	cs.chunks[0].initialize(part, t)
+	if cs.chunks[0] != nil {
```
imakhlin: same here - use `cs.AggregatesOnly`

imakhlin:
Add necessary unit & integration tests

pkg/partmgr/partmgr.go
```diff
@@ -342,8 +342,12 @@ func (p *DBPartition) GetShardingKeys(name string) []string {
 }
 
 // Return the full path to the specified metric item
-func (p *DBPartition) GetMetricPath(name string, hash uint64) string {
-	return fmt.Sprintf("%s%s_%x.%016x", p.path, name, int(hash%uint64(p.GetHashingBuckets())), hash)
+func (p *DBPartition) GetMetricPath(name string, hash uint64, isAggr bool) string {
+	agg := ""
+	if isAggr {
+		agg = "agg/"
```
imakhlin: 1. It would be nice to have a design page that describe all aspects of pre-aggr feature.
2. Adding new folder to the table hierarchy has impact on the **Delete**/Retention/Rollup/etc. Now, Delete table command should take care of the nested folders and their content.

imakhlin:
This PR could not be tested without implementing Query & Delete, therefore I would suggest postpone it and implement the dependant components first.

pkg/appender/store.go
```diff
@@ -326,8 +339,10 @@ func (cs *chunkStore) writeChunks(mc *MetricsCache, metric *MetricState) (hasPen
 			// Add a value to the aggregates list
 			cs.aggrList.Aggregate(sampleTime, cs.pending[pendingSampleIndex].v)
 
-			// Add a value to the compressed raw-values chunk
-			activeChunk.appendAttr(sampleTime, cs.pending[pendingSampleIndex].v)
+			if activeChunk != nil {
```
talIguaz: even though it's basically the same, i would consider changing the condition to `!cs.isAggr()` because this is the real case you want to append stuff to the chunk. so it might be more clear.

pkg/appender/store.go
```diff
@@ -352,7 +367,7 @@ func (cs *chunkStore) writeChunks(mc *MetricsCache, metric *MetricState) (hasPen
 
 			// If the next item is in a new chunk, generate an expression and
 			// initialize the new chunk
-			if !activeChunk.inRange(nextT) {
+			if activeChunk != nil && !activeChunk.inRange(nextT) {
```
talIguaz: same regarding `!cs.isAggr()`

pkg/utils/labels.go
```diff
@@ -40,13 +40,32 @@ type Label struct {
 	Name, Value string
 }
 
+type LabelIfc interface {
```
talIguaz: Seems like nobody uses it

PR #224, opened 2018-12-03
fix /series

pavius:
When `/series` is called in Prometheus, a `nil` params structure is passed. This PR will create a params structure with default settings in such a case.


gshatz:
This fix will work on small tables but for big ones it will take very long time to return since it literally collect ALL the data of the table.
I'm leaving this PR so @levrado will be able to continue and we will open another PR to collect only the lset from the first partition (to avoid collecting all blobs and arrays)

PR #233, opened 2018-12-11
Variant encoder

talIguaz:


pkg/appender/appender.go
```diff
@@ -217,6 +218,10 @@ func (mc *MetricsCache) Add(lset utils.LabelsIfc, t int64, v interface{}) (uint6
 
 	if !ok {
 		metric = &MetricState{Lset: lset, key: key, name: name, hash: hash}
+		// if the (first) value is not float, use variant encoding, TODO: test w schema
+		if _, ok := v.(float64); !ok {
```
imakhlin: Maybe you should take care of `NaN`. With `NaN` you can't determine the type, so it's better to skip it

pkg/aggregate/aggregate.go
```diff
@@ -194,7 +194,10 @@ type AggregatesList []Aggregate
 
 // Append a value to all aggregates
 func (a AggregatesList) Aggregate(t int64, val interface{}) {
-	v := val.(float64)
+	v, ok := val.(float64)
```
imakhlin: If you want to skip the `variants` you should call aggregate at all

pkg/appender/store.go
```diff
@@ -224,7 +223,7 @@ func (cs *chunkStore) Append(t int64, v interface{}) {
 }
 
 // Return current, previous, or create new  chunk based on sample time
-func (cs *chunkStore) chunkByTime(t int64) *attrAppender {
+func (cs *chunkStore) chunkByTime(t int64, isVar bool) *attrAppender {
```
imakhlin: minor: use self explaining names

pkg/chunkenc/chunk.go
```diff
@@ -44,6 +43,8 @@ func (e Encoding) String() string {
 		return "none"
 	case EncXOR:
 		return "XOR"
+	case EncVar:
```
imakhlin: cosmetic: `EncVariant`

pkg/chunkenc/vartype.go
```diff
@@ -0,0 +1,257 @@
...
+
+type VarChunk struct {
+	logger logger.Logger
+
+	b       []byte
+	samples uint16
+	offset  int
+}
+
+// NewVarChunk returns a new chunk with variant encoding of the given size.
```
imakhlin: In comment you say "given size" please correct comment or code
Shouldn't chunk size be configurable?

pkg/chunkenc/vartype.go
```diff
@@ -0,0 +1,257 @@
...
+type varAppender struct {
+	logger logger.Logger
+
+	c       *VarChunk
+	samples *uint16
+	t       int64
+}
+
+func (a *varAppender) Encoding() Encoding {
+	return a.Chunk().Encoding()
```
imakhlin: > a.c.Encoding()

pkg/chunkenc/vartype.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+
+		a.appendWithUint(t, varTypeFloat64, math.Float64bits(val))
+
+	case string:
+		val := []byte(v.(string))
+		a.appendWithValue(t, varTypeString, val)
+
+	default:
+		a.logger.Error("unsupported type %T!\n", vType)
```
imakhlin: > yields **syntax error: unexpected type, expecting expression**
you should use `v` instead

pkg/chunkenc/vartype_test.go
```diff
@@ -0,0 +1,54 @@
...
+	suite.Require().Nil(err)
+
+	list := []string{"abc", "", "123456"}
+	t0 := time.Now().UnixNano() / 1000
+
+	for i, s := range list {
+		t := t0 + int64(i*1000)
+		appender.Append(t, s)
+		b := chunk.Bytes()
+		fmt.Println(t, s, len(b))
```
imakhlin: Do you really need those prints during the test?

pkg/chunkenc/vartype_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/nuclio/zap"
+	"github.com/stretchr/testify/suite"
+)
+
+type testVarEncoderSuite struct {
+	suite.Suite
+}
+
+func (suite *testVarEncoderSuite) TestStringEnc() {
```
imakhlin: What about rest data types?
> 	varTypeNil     byte = 0
	varTypeBlob    byte = 1
	varTypeString  byte = 2
	varTypeBool    byte = 3
	varTypeFloat32 byte = 4
	varTypeFloat64 byte = 5
	varTypeInt8    byte = 8
	varTypeInt16   byte = 9
	varTypeInt32   byte = 10
	varTypeInt64   byte = 11

pkg/aggregate/aggregate.go
talIguaz: It happens automatically when ingesting data. Now if you have one tsdb with 2 metrics one of them is normal (like cpu) the other is of variant type (like log).
you created the table with default aggregates that are being aggregated on ingest for every metric in the DB so we need to ignore non-floats.
At least until a better solution.

pkg/appender/appender.go
talIguaz: Discussed in person

talIguaz:
As discussed in person, will extend the support to all other types (other than `string`) in another PR

PR #238, opened 2018-12-17
Support session key

gshatz:
1. Add support for passing session-key by command (-k) or by config (sessionKey) in order to authenticate the requests to nginx (needed for 2.0). Session-key will take precedence on user/password (in case both will be passed)

2. Vendorize new v3io-http-go with the ability to pass the session-key as header for the request (PR: https://github.com/v3io/v3io-go-http/pull/13)


gshatz:
Removed redundant files from vendorized v3io-http-go.
As for Integration test, this is tricky because there is no valid session-key by default on any env (like the default user/password we have) so before we can test this with integration need to create the ability to create a session-key.
Unitest exists

PR #240, opened 2018-12-20
Query multiple metrics in CLI

talIguaz:


gtopper:
Looks good, just change the docu accordingly :)

PR #246, opened 2018-12-27
Filter when querying unique labelsets

talIguaz:


gshatz:
please add integration test

PR #258, opened 2019-01-02
Use pre-aggregations for groupBy queries.

gtopper:


pkg/pquerier/select.go
```diff
@@ -172,19 +172,44 @@ func (queryCtx *selectQueryContext) queryPartition(partition *partmgr.DBPartitio
...
-			newQuery.preAggregated = aggregationParams.CanAggregate(partition.AggrType())
-			if newQuery.preAggregated || !queryCtx.queryParams.disableClientAggr {
+			newQuery.useServerSideAggregates = aggregationParams.CanAggregate(partition.AggrType())
+			if newQuery.useServerSideAggregates || !queryCtx.queryParams.disableClientAggr {
 				newQuery.aggregationParams = aggregationParams
 			}
 		}
 
-		err = newQuery.getItems(queryCtx, metric, requestAggregatesAndRaw)
+		groupBy := queryCtx.groupBy(partition)
```
talIguaz: i would change the name of the function&variable to something more clear, since `groupBy` by default is the "group by" statement you pass in the select but here it's a bit different meaning

pkg/pquerier/select.go
```diff
@@ -172,19 +172,44 @@ func (queryCtx *selectQueryContext) queryPartition(partition *partmgr.DBPartitio
...
+		err = newQuery.getItems(queryCtx, metric, groupBy, requestAggregatesAndRaw)
 		queries = append(queries, newQuery)
 	}
 
 	return queries, err
 }
 
+func (queryCtx *selectQueryContext) groupBy(partition *partmgr.DBPartition) []string {
```
talIguaz: i'd move this to `partQuery`

pkg/pquerier/select.go
```diff
@@ -408,27 +433,31 @@ type partQuery struct {
...
+	chunkTime               int64
+	useServerSideAggregates bool
+	aggregationParams       *aggregate.AggregationParams
 }
 
-func (query *partQuery) getItems(ctx *selectQueryContext, name string, aggregatesAndChunk bool) error {
+func (query *partQuery) getItems(ctx *selectQueryContext, name string, groupBy []string, aggregatesAndChunk bool) error {
 
 	path := query.partition.GetTablePath()
+	if query.useServerSideAggregates && !aggregatesAndChunk && len(groupBy) > 0 {
```
talIguaz: a bit petty but, i would move the `len(groupBy) > 0` part to be the first condition in the `if` because first of all, this is the real essence of the condition. Plus because we're such a high performance product, and mostly this condition would "fail" the if it should be first 😅

pkg/pquerier/select.go
gtopper: But it needs the `queryParams`.

pkg/pquerier/select.go
gtopper: Better yet - I'll avoid resolving the groupBy labels altogether in that case.

pkg/pquerier/select.go
gtopper: `groupByLabels`?

talIguaz: `parsePreAggregatedLabelGroups` ?
and the variable `preAggregatedLabelGroups`

pkg/pquerier/select.go
talIguaz: yea well the `getItems` gets the selectQueryContext as parameter and seems like this logic should be there.. but idk

PR #267, opened 2019-01-08
Change time column type to time.Time

talIguaz:


pkg/pquerier/frames.go
```diff
@@ -681,14 +681,6 @@ func (c *virtualColumn) StringAt(i int) (string, error) {
...
-		return 0, fmt.Errorf("index %d out of bounds [0:%d]", i, c.size)
-	}
-
-	value, err := c.function(c.dependantColumns, i)
-	if err != nil {
-		return 0, err
-	}
-	return value.(int64), nil
+func (c *virtualColumn) TimeAt(i int) (time.Time, error) {
+	return time.Unix(0, 0), errors.New("aggregated column does not support time type")
```
gtopper: I guess the impl that was there didn't work?

PR #271, opened 2019-01-09
Consider alias name when returning series

talIguaz:


gshatz:
@talIguaz do we have test for the use of alias if exist?

PR #280, opened 2019-01-14
Fix cross series aggregates logic

talIguaz:


pkg/pquerier/collector.go
```diff
@@ -72,16 +74,25 @@ func mainCollector(ctx *selectQueryContext, responseChannel chan *qryResults) {
...
-			} else if res.IsClientAggregates() {
-				aggregateClientAggregates(ctx, res)
+			lsetAttr, _ := res.fields[config.LabelSetAttrName].(string)
+			lset, _ := utils.LabelsFromString(lsetAttr)
+			currentResultHash := lset.Hash()
+
+			// Aggregating cross series aggregates, only supported over raw data.
+			if ctx.isCrossSeriesAggregate {
+				lastTimePerMetric[currentResultHash], lastValuePerMetric[currentResultHash], _ = aggregateClientAggregatesCrossSeries(ctx, res, lastTimePerMetric[currentResultHash], lastValuePerMetric[currentResultHash])
+				// Aggregating over time aggregates
```
gtopper: What's this comment doing here at the end of a block?

pkg/pquerier/collector.go
```diff
@@ -204,3 +215,57 @@ func downsampleRawData(ctx *selectQueryContext, res *qryResults,
 
 	return lastT, lastV, nil
 }
+
+func aggregateClientAggregatesCrossSeries(ctx *selectQueryContext, res *qryResults, previousPartitionLastTime int64, previousPartitionLastValue float64) (int64, float64, error) {
+	ctx.logger.Debug("using Client Aggregates Collector for metric %v", res.name)
+	it := newRawChunkIterator(res, nil).(*rawChunkIterator)
```
gtopper: Pass the logger in? `newRawChunkIterator` should already return a `rawChunkIterator`.

pkg/pquerier/collector.go
talIguaz: it's supposed to be for the `else`, but for some reason this is how go formatted it

pkg/pquerier/collector.go
```diff
@@ -204,3 +215,57 @@ func downsampleRawData(ctx *selectQueryContext, res *qryResults,
...
+
+	for currBucket := previousPartitionEndBucket; currBucket < maxBucketForPartition; currBucket++ {
+		currBucketTime := int64(currBucket)*ctx.queryParams.Step + ctx.queryParams.From
+
+		if it.Seek(currBucketTime) {
+			t, v := it.At()
+			if t == currBucketTime {
+				for _, col := range res.frame.columns {
+					if col.GetColumnSpec().metric == res.name {
+						col.SetDataAt(currBucket, v)
```
gtopper: If we're going to ignore errors, we should use `_ = ` notation to explicitly do so (I'm surprised the linter is ok with this - I think that's not the case in Nuclio).

pkg/pquerier/collector.go
gtopper: So move it inside the else block?

pkg/pquerier/query_integration_test.go
```diff
@@ -1888,6 +1888,430 @@ func (suite *testQuerySuite) toMillis(date string) int64 {
...
+	adapter, err := tsdb.NewV3ioAdapter(suite.v3ioConfig, nil, nil)
+	suite.Require().NoError(err, "failed to create v3io adapter")
+
+	labels1 := utils.LabelsFromStringList("os", "linux")
+	labels2 := utils.LabelsFromStringList("os", "mac")
+	numberOfEvents := 10
+	eventsInterval := 60 * 1000
+	baseTime := tsdbtest.NanosToMillis(time.Now().UnixNano()) - int64(numberOfEvents*eventsInterval)
+
+	fmt.Println("the base time is: ", baseTime)
```
gtopper: ⚡️

pkg/pquerier/query_integration_test.go
```diff
@@ -1888,6 +1888,430 @@ func (suite *testQuerySuite) toMillis(date string) int64 {
...
```
gtopper: ⚡️

pkg/pquerier/query_integration_test.go
```diff
@@ -1888,6 +1888,430 @@ func (suite *testQuerySuite) toMillis(date string) int64 {
...
```
gtopper: ⚡️

pkg/pquerier/collector.go
```diff
@@ -204,3 +215,57 @@ func downsampleRawData(ctx *selectQueryContext, res *qryResults,
...
+				// In case it's the first point in the partition use the last point of the previous partition for the interpolation
+				if prevT == 0 {
+					prevT = previousPartitionLastTime
+					prevV = previousPartitionLastValue
+				}
+
+				for _, col := range res.frame.columns {
+					if col.GetColumnSpec().metric == res.name {
+						interpolatedT, interpolatedV := col.GetInterpolationFunction()(prevT, t, currBucketTime, prevV, v)
+						if interpolatedT != 0 && interpolatedV != 0 {
```
gtopper: Bug 🐛🙈

pkg/pquerier/collector.go
```diff
@@ -110,14 +121,14 @@ func rawCollector(ctx *selectQueryContext, res *qryResults) {
 
 func aggregateClientAggregates(ctx *selectQueryContext, res *qryResults) {
 	ctx.logger.Debug("using Client Aggregates Collector for metric %v", res.name)
-	it := newRawChunkIterator(res, nil)
+	it := newRawChunkIterator(res, ctx.logger.GetChild("rawChunkIterator"))
```
gtopper: This (`GetChild`) is generally done inside the constructor. See [nuclio](https://github.com/nuclio/nuclio/blob/f2ebdd95e95c4a1cb24fcb14e18ea88cd9093eba/pkg/processor/trigger/kinesis/trigger.go#L42) for an example.

pkg/pquerier/collector.go
```diff
@@ -165,7 +176,7 @@ func downsampleRawData(ctx *selectQueryContext, res *qryResults,
 
 	var lastT int64
 	var lastV float64
```
gtopper: Would be nice to move these as well.

pkg/pquerier/interpolate.go
```diff
@@ -56,46 +73,55 @@ func StrToInterpolateType(str string) (InterpolationType, error) {
...
-}
-
-// linear estimator (smooth graph)
-func projectLinear(tprev, tnext, tseek int64, vprev, vnext float64) (int64, float64) {
-	if math.IsNaN(vprev) || math.IsNaN(vnext) {
-		return tseek, math.NaN()
+func validateTolerance(tolerance, tseek, tprev, tnext int64) bool {
+	// If previous point is too far behind for interpolation or the next point is too far ahead then return NaN
+	if (tprev != 0 && tseek-tprev > tolerance) || tnext-tseek > tolerance {
+		return false
```
gtopper: The comment seems to be wrong.

PR #281, opened 2019-01-14
Release processed chunks in querier

gshatz:
and validate getItems results to avoid infinite loop

### DIFF
--- a/pkg/utils/asynciter.go
+++ b/pkg/utils/asynciter.go
@@ -167,7 +167,7 @@ func (ic *AsyncItemsCursor) NextItem() (v3io.Item, error) {
 	}
 
 	getItemsResp := resp.Output.(*v3io.GetItemsOutput)
-	resp.Output = nil;
+	resp.Output = nil
 
 	// set the cursor items and reset the item index
 	ic.items = getItemsResp.Items
--- a/pkg/utils/asynciter.go
+++ b/pkg/utils/asynciter.go
@@ -167,6 +167,7 @@ func (ic *AsyncItemsCursor) NextItem() (v3io.Item, error) {
 	}
 
 	getItemsResp := resp.Output.(*v3io.GetItemsOutput)
+	resp.Output = nil;
 
 	// set the cursor items and reset the item index
 	ic.items = getItemsResp.Items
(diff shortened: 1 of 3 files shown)


pkg/utils/asynciter.go
```diff
@@ -167,6 +167,7 @@ func (ic *AsyncItemsCursor) NextItem() (v3io.Item, error) {
 	}
 
 	getItemsResp := resp.Output.(*v3io.GetItemsOutput)
+	resp.Output = nil
```
talIguaz: Not sure it is necessary, look at `asynciter.go:156` it releases the responses on defer

vendor/github.com/v3io/v3io-go-http/synccontainer.go
```diff
@@ -236,6 +236,15 @@ func (sc *SyncContainer) GetItems(input *GetItemsInput) (*Response, error) {
 		return nil, err
 	}
 
+	//validate range scan loop
```
talIguaz: Open a PR to v3io-go-http and then we will re-vendor with the latest v3io-go-http version.

gshatz: already did - this is the only change

pkg/utils/asynciter.go
gshatz: Response.Release() only returns the fastHttpResponse body back to the pool, it doesn't release the base64 decoded blobs

vendor/github.com/v3io/v3io-go-http/synccontainer.go
gtopper: This diff doesn't match https://github.com/v3io/v3io-go-http/pull/14

pkg/utils/asynciter.go
gtopper: This line does nothing :)

gtopper: Ok, I found the actual leak. It's the `defer resp.Release()`, which is inappropriate in case of recursion, and causes both the fasthttp response and the wrapper to be preserved until the recursion is done. Instead, it should be released __before__ the recursive call.

gtopper: https://github.com/v3io/v3io-tsdb/pull/284

PR #282, opened 2019-01-15
Reduce object size to 1m

gshatz:


talIguaz:
@gshatz  Actually why did you change it in the first place? this option is configurable so for 1.9 stuff where we still have the 2MB limitation you can just pass a v3io.yaml with a different max partition size.

PR #283, opened 2019-01-15
Fix RawIterator's peakBack

talIguaz:
…zero

### DIFF
--- a/pkg/pquerier/chunkIterator.go
+++ b/pkg/pquerier/chunkIterator.go
@@ -11,7 +11,7 @@ import (
 )
 
 // Chunk-list series iterator
-type rawChunkIterator struct {
+type RawChunkIterator struct {
 	mint, maxt int64
 
 	chunks   []chunkenc.Chunk
@@ -33,7 +33,7 @@ func newRawChunkIterator(queryResult *qryResults, log logger.Logger) utils.Serie
 		maxt = int64(maxTime.(int))
 	}
 
-	newIterator := rawChunkIterator{
+	newIterator := RawChunkIterator{
 		mint: queryResult.query.mint, maxt: maxt, log: log, encoding: queryResult.encoding}
 
 	newIterator.AddChunks(queryResult)
@@ -48,7 +48,7 @@ func newRawChunkIterator(queryResult *qryResults, log logger.Logger) utils.Serie
 }
 
 // Advance the iterator to the specified chunk and time
-func (it *rawChunkIterator) Seek(t int64) bool {
+func (it *RawChunkIterator) Seek(t int64) bool {
 
 	// Seek time is after the item's end time (maxt)
 	if t > it.maxt {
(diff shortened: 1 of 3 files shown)


pkg/pquerier/chunkIterator.go
```diff
@@ -99,15 +99,15 @@ func (it *rawChunkIterator) Seek(t int64) bool {
 	}
 }
 
-func (it *rawChunkIterator) updatePrevPoint() {
+func (it *RawChunkIterator) updatePrevPoint() {
 	t, v := it.At()
-	if t != 0 && v != 0 {
+	if !(t == 0 && v == 0) {
 		it.prevT, it.prevV = t, v
```
gtopper: Or `t != 0 || v != 0`? 😆

pkg/pquerier/chunkIterator_test.go
```diff
@@ -0,0 +1,111 @@
...
+type testRawChunkIterSuite struct {
+	suite.Suite
+	v3ioConfig     *config.V3ioConfig
+	suiteTimestamp int64
+}
+
+func (suite *testRawChunkIterSuite) SetupSuite() {
+	v3ioConfig, err := tsdbtest.LoadV3ioConfig()
+	if err != nil {
+		suite.T().Fatalf("unable to load configuration. Error: %v", err)
```
gtopper: `Require().NoError()`?

pkg/pquerier/chunkIterator_test.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	suite.v3ioConfig = v3ioConfig
+	suite.suiteTimestamp = time.Now().Unix()
```
gtopper: Can we use a constant time?

talIguaz: This is for tsdb table naming format.
But i did change the actual `baseTime` i use in the test to be constant

gtopper: Ok, using a changing location makes sense of course, so tests are reliably repeatable.

pkg/pquerier/chunkIterator_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (suite *testRawChunkIterSuite) TestRawChunkIteratorWithZeroValue() {
+	adapter, err := tsdb.NewV3ioAdapter(suite.v3ioConfig, nil, nil)
+	suite.Require().NoError(err)
+
+	labels1 := utils.LabelsFromStringList("os", "linux")
+	numberOfEvents := 10
+	eventsInterval := 60 * 1000
+	ingestData := []tsdbtest.DataPoint{{baseTestTime, 10},
+		{int64(baseTestTime + tsdbtest.MinuteInMillis), 0},
```
gtopper: Looks like an unnecessary conversion given that other entries don't have it.

pkg/pquerier/chunkIterator_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+			if index > 0 {
+				suite.Require().Equal(ingestData[index-1].Time, prevT, "current time does not match")
+				suite.Require().Equal(ingestData[index-1].Value, prevV, "current value does not match")
+			}
+			index++
+		}
+	}
+
+	assert.Equal(suite.T(), 1, seriesCount, "series count didn't match expected")
```
gtopper: Use `suite.Require().Equal(...)`?

PR #284, opened 2019-01-15
Fixed memory leak caused by a mix of recursion and defer.

gtopper:
https://github.com/v3io/v3io-tsdb/pull/284/files?w=1

### DIFF
--- a/pkg/utils/asynciter.go
+++ b/pkg/utils/asynciter.go
@@ -151,17 +151,14 @@ func (ic *AsyncItemsCursor) NextItem() (v3io.Item, error) {
 			return nil, nil
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/utils/asynciter.go
```diff
@@ -132,25 +132,36 @@ func (ic *AsyncItemsCursor) Next() bool {
...
+		if err != nil {
+			return nil, err
+		}
+		if res != nil {
+			return res, nil
+		}
 	}
+}
 
+func (ic *AsyncItemsCursor) processResponse() (v3io.Item, error) {
```
talIguaz: the `v3io.Item` return argument is redundant, you always return nil..

PR #285, opened 2019-01-16
Change interpolation SQL syntax

talIguaz:
change interpolation syntax from `select linear.cpu` to `select linear(cpu, '1h')`


talIguaz:
The SQL query syntax is not documented at all. so there is nothing to change

PR #289, opened 2019-01-20
Support cross-label aggregates in create.

gtopper:


pkg/tsdbctl/create.go
```diff
@@ -62,6 +63,8 @@ func newCreateCommandeer(rootCommandeer *RootCommandeer) *createCommandeer {
 
 	cmd.Flags().StringVarP(&commandeer.defaultRollups, "aggregates", "a", "",
 		"Default aggregates to calculate in real time during\nthe samples ingestion, as a comma-separated list of\nsupported aggregation functions - count | avg | sum |\nmin | max | stddev | stdvar | last | rate.\nExample: \"sum,avg,max\".")
+	cmd.Flags().StringVarP(&commandeer.crossLabelSets, "cross-label", "l", "",
+		"Label sets for which cross-label pre-aggregations should be created. Must be used in conjunction with -a.\nExample: \"label1,label2.label3.\".")
```
gshatz: I find the dot non-intuitive. WDYT of "label1,label2||label3" ?

pkg/aggregate/aggregate.go
```diff
@@ -129,6 +129,25 @@ func RawAggregatesToStringList(aggregates string) ([]string, error) {
 	return list, nil
 }
 
+func ParseCrossLabelSets(str string) [][]string {
+	var res [][]string
+	labelSetStrings := strings.Split(str, ".")
```
gshatz: need to consider the dot here with another delimiter...

gtopper: Agreed to use semicolon instead of dot.

pkg/tsdbctl/create.go
gtopper: Agreed to use semicolon instead of dot.

README.md
```diff
@@ -81,8 +81,9 @@ A user can run the CLI to add (append) or query the DB, to use the CLI, build th
 it has built-in help, see the following add/query examples:
 
 ```
-	# create a DB with expected ingestion rate of one sample per second and some aggregates (at 30 min interval) 
-	tsdbctl create -t <table> --ingestion-rate 1/s -a count,sum,max -i 30m
+	# create a DB with expected ingestion rate of one sample per second and some aggregates (at 30 min interval)
+	# and cross-label aggregates for "host" 
+	tsdbctl create -t <table> --ingestion-rate 1/s -a count,sum,max -i 30m -l label1
```
gshatz: change label1 to host

gshatz:
Need to add tests

PR #296, opened 2019-02-11
TSDB-78: Sanitise configuration string representation

imakhlin:
- implement toString method for V3ioConfig that hides sensitive information like password and access key


imakhlin:
@talIguaz please make your tests deterministic , so it won't break CI 
;-)

PR #297, opened 2019-02-11
TSDB-81: take labelValues from last (full) partition instead of first partition

gshatz:


pkg/querier/querier.go
```diff
@@ -298,17 +298,24 @@ func (q *V3ioQuerier) getLabelValues(labelKey string) ([]string, error) {
...
 
 	// If there are no partitions yet, there are no labels
-	if len(partitionPaths) == 0 {
+	if numPartitions == 0 {
 		return nil, nil
 	}
 
+	//take the last FULL partition (unless there is only 1-2 partitions)
+	var partitionIndex = numPartitions - 1
+	if numPartitions > 2 {
```
talIguaz: `numPartitions > 1`

gshatz: if I have only 2 than I don't know which one of them is full... only when I have 3 and more I can assume the last-1 is probably full

PR #299, opened 2019-02-13
Support delete NaNs

talIguaz:


pkg/pquerier/collector.go
```diff
@@ -195,6 +197,7 @@ func downsampleRawData(ctx *selectQueryContext, res *qryResults,
 				} else {
 					_, interpolatedV := interpolateFunc(prevT, t, currBucketTime, prevV, v)
 					col.SetDataAt(currBucket, interpolatedV)
+					res.frame.updateHasValue(currBucket)
```
imakhlin: use same terminology `bucket` or `cell`

pkg/pquerier/frames.go
```diff
@@ -105,6 +108,7 @@ func NewDataFrame(columnsSpec []columnMeta, indexColumn Column, lset utils.Label
 		df.columns = make([]Column, 0, numOfColumns)
 		df.metricToCountColumn = map[string]Column{}
 		df.metrics = map[string]struct{}{}
+		df.nonEmptyRowsIndicator = make([]bool, columnSize)
```
imakhlin: you can avoid this if you'll maintain simple int counter in a column for num of non empty rows

imakhlin: `col.SetDataAt(..)` could increment/decrement the column counter

pkg/pquerier/select.go
```diff
@@ -396,11 +394,8 @@ func (queryCtx *selectQueryContext) getOrCreateTimeColumn() Column {
 	if queryCtx.isRawQuery() {
 		return nil
 	}
-	if queryCtx.timeColumn == nil {
-		queryCtx.timeColumn = queryCtx.generateTimeColumn()
-	}
 
-	return queryCtx.timeColumn
+	return queryCtx.generateTimeColumn()
```
imakhlin: How about `ctx.getOrCreateTimeColumn()`?

pkg/pquerier/frames.go
```diff
@@ -210,15 +214,22 @@ type dataFrame struct {
...
+	columns               []Column
+	index                 Column
+	columnByName          map[string]int // name -> index in columns
+	nonEmptyRowsIndicator []bool
 
 	metrics             map[string]struct{}
 	metricToCountColumn map[string]Column
 }
 
+func (d *dataFrame) updateHasValue(index int) {
```
imakhlin: Looks like you never use it ...

pkg/pquerier/frames.go
```diff
@@ -253,6 +264,20 @@ func (d *dataFrame) addMetricFromTemplate(metricName string, columnSize int, use
 	return nil
 }
 
+func (d *dataFrame) setDataAt(columnName string, index int, value interface{}) error {
+	colIndex, ok := d.columnByName[columnName]
+	if !ok {
+		return fmt.Errorf("no such column %v", columnName)
+	}
+	col := d.columns[colIndex]
+	err := col.SetDataAt(index, value)
```
imakhlin: Can value be considered as empty? Value like NaN or empty string

talIguaz: Currently i removed all places where i "set" NaN to a certain index so i only set "real" values.
Regarding string, because we only support raw data for string type there are no "empty strings"

pkg/pquerier/frames.go
```diff
@@ -331,27 +350,48 @@ func (d *dataFrame) TimeSeries(i int) (utils.Series, error) {
...
 	for _, col := range d.columns {
 		switch col.(type) {
 		case *dataColumn:
 			err = col.finish()
 		case *ConcreteColumn:
 			err = col.finish()
+			if columnSize == 0 {
+				columnSize = col.FramesColumn().Len()
+			} else if columnSize != col.FramesColumn().Len() {
+				return fmt.Errorf("columns length mismath %v!=%v col=%v", columnSize, col.FramesColumn().Len(), col.Name())
```
imakhlin: Typo "mismatch"

PR #301, opened 2019-02-17
TSDB-82: Change it-tests to use const time + server aggr bug fix

talIguaz:


gshatz:
@talIguaz please add tests to cover http://jira.iguazeng.com:8080/browse/TSDB-82

PR #319, opened 2019-03-06
TSDB-93

talIguaz:


pkg/pquerier/frames.go
```diff
@@ -355,14 +355,31 @@ func (d *dataFrame) TimeSeries(i int) (utils.Series, error) {
...
+		hasData := d.nonEmptyRowsIndicators[i]
 		if !hasData {
 			for _, col := range d.columns {
 				_ = col.Delete(i)
 			}
 			_ = d.index.Delete(i)
+		} else {
+			for _, col := range d.columns {
+				switch currCol := col.(type) {
+				case *dataColumn:
```
gshatz: I see both cases are equaly treated, why not do "case *dataColumn, *ConcreteColumn :" ?

pkg/pquerier/query_integration_test.go
```diff
@@ -3075,6 +3075,159 @@ func (suite *testQuerySuite) TestDifferentMetricsInDifferentPartitions() {
...
+
+				suite.Require().NoError(err)
+				if !(math.IsNaN(currVal) && math.IsNaN(expected[currName][i])) {
+					suite.Require().Equal(expected[currName][i], currVal)
+				}
+			}
+		}
+	}
+
+	assert.Equal(suite.T(), 1, dataFrameCount, "series count didn't match expected")
```
gtopper: `suite.Require().Equal`?

PR #323, opened 2019-03-11
IG-11142 Support aggregation window

talIguaz:


pkg/tsdbctl/query.go
```diff
@@ -186,6 +179,10 @@ func (qc *queryCommandeer) newQuery(from, to, step int64) error {
 	if err != nil {
 		return errors.Wrap(err, "Failed to initialize the Querier object.")
 	}
+	aggregationWindow, err := utils.Str2duration(qc.aggregationWindow)
+	if err != nil {
+		return err
```
gtopper: We should probably say the issue was with the aggregation window in the error message (i.e. wrap the error).

pkg/tsdbctl/query.go
```diff
@@ -100,13 +99,8 @@ Arguments:
...
-	// The default aggregates list for an overlapping-windows query is "avg",
-	// provided the TSDB instance has the "count" and "sum" aggregates, which
-	// make up the "avg" aggregate; ("count" is added automatically when adding
-	// any other aggregate). However, it was decided that documenting this
-	// would over complicate the documentation.
+	cmd.Flags().StringVarP(&commandeer.aggregationWindow, "aggregation-window", "w", "",
+		"Sliding window to calculate the aggregation for every step. Examples: \"1h\"; \"150m\". (default = 0)")
```
gtopper: The part about every step is not really clear. What does it mean for the window to be 0 (the default)?

pkg/pquerier/collector.go
```diff
@@ -263,3 +273,69 @@ func aggregateClientAggregatesCrossSeries(ctx *selectQueryContext, res *qryResul
...
+}
+
+func windowAggregation(ctx *selectQueryContext, res *qryResults, t int64, v float64) {
+	currentCell := (t - ctx.queryParams.From) / res.query.aggregationParams.Interval
+	if (t-ctx.queryParams.From)%res.query.aggregationParams.Interval > 0 {
+		currentCell++
+	}
+
+	if res.query.aggregationParams.GetAggregationWindow() > res.query.aggregationParams.Interval {
+		numEffectedCells := (t + res.query.aggregationParams.GetAggregationWindow() - ctx.queryParams.From - currentCell*res.query.aggregationParams.Interval) / res.query.aggregationParams.Interval
```
gtopper: Simplify? Also, could help to alias stuff like `res.query.aggregationParams.GetAggregationWindow()` to form a more readable expression.

pkg/pquerier/collector.go
```diff
@@ -263,3 +273,69 @@ func aggregateClientAggregatesCrossSeries(ctx *selectQueryContext, res *qryResul
...
+		if res.query.aggregationParams.GetAggregationWindow()%res.query.aggregationParams.Interval > 0 {
+			numEffectedCells++
```
gtopper: Effected -> Affected 😛

pkg/pquerier/collector.go
```diff
@@ -263,3 +273,67 @@ func aggregateClientAggregatesCrossSeries(ctx *selectQueryContext, res *qryResul
 	lastT, lastV := it.At()
 	return lastT, lastV, nil
 }
+
+func intervalAggregation(ctx *selectQueryContext, res *qryResults, t int64, v float64) {
+	currentCell := (t - ctx.queryParams.From) / res.query.aggregationParams.Interval
```
imakhlin: do not repeat - move to a function

pkg/pquerier/collector.go
```diff
@@ -263,3 +273,67 @@ func aggregateClientAggregatesCrossSeries(ctx *selectQueryContext, res *qryResul
...
+	if (t-ctx.queryParams.From)%res.query.aggregationParams.Interval > 0 {
+		currentCell++
+	}
+
+	if res.query.aggregationParams.GetAggregationWindow() > res.query.aggregationParams.Interval {
+		currentCellTime := ctx.queryParams.From + currentCell*res.query.aggregationParams.Interval
+		maximumAffectedTime := t + res.query.aggregationParams.GetAggregationWindow()
+		numAffectedCells := (maximumAffectedTime-currentCellTime)/res.query.aggregationParams.Interval + 1 // +1 to include the current cell
+
+		for i := int64(0); i < numAffectedCells; i++ {
```
imakhlin: why to cast here? cast `numAffectedCells` to int and that's it. Right?

pkg/pquerier/collector.go
```diff
@@ -263,3 +273,67 @@ func aggregateClientAggregatesCrossSeries(ctx *selectQueryContext, res *qryResul
...
+			for _, col := range res.frame.columns {
+				if col.GetColumnSpec().metric == res.name {
+					_ = res.frame.setDataAt(col.Name(), int(currentCell+i), v)
+				}
+			}
+		}
+	} else if res.query.aggregationParams.GetAggregationWindow() < res.query.aggregationParams.Interval {
```
imakhlin: again.. store intermediate results...

pkg/pquerier/collector.go
```diff
@@ -263,3 +273,67 @@ func aggregateClientAggregatesCrossSeries(ctx *selectQueryContext, res *qryResul
...
+		if t+res.query.aggregationParams.GetAggregationWindow() >= ctx.queryParams.From+currentCell*res.query.aggregationParams.Interval {
+			for _, col := range res.frame.columns {
+				if col.GetColumnSpec().metric == res.name {
+					_ = res.frame.setDataAt(col.Name(), int(currentCell), v)
+				}
+			}
+		}
+	} else {
+		for _, col := range res.frame.columns {
```
imakhlin: isn't this the same case as `intervalAggregation()`?

pkg/pquerier/query_integration_test.go
```diff
@@ -3286,6 +3286,460 @@ func (suite *testQuerySuite) TestSelectDataframeDaownsampleMetricsHaveBigGaps()
 	suite.Require().Equal(1, dataFrameCount, "series count didn't match expected")
 }
 
+func (suite *testQuerySuite) TestClientWindowedAggregationWindowBiggerThanStep() {
```
imakhlin: This file contains TOO many tests. Consider to break it into categories.

pkg/pquerier/query_integration_test.go
```diff
@@ -3286,6 +3286,460 @@ func (suite *testQuerySuite) TestSelectDataframeDaownsampleMetricsHaveBigGaps()
...
+	labels1 := utils.LabelsFromStringList("os", "linux")
+	numberOfEvents := 10
+
+	var ingestedData []tsdbtest.DataPoint
+
+	for i := 0; i < numberOfEvents; i++ {
+		ingestedData = append(ingestedData, tsdbtest.DataPoint{Time: suite.basicQueryTime + int64(i)*tsdbtest.MinuteInMillis, Value: 10 * float64(i)})
+	}
+
+	testParams := tsdbtest.NewTestParams(suite.T(),
```
imakhlin: Can you reuse all these "before" test blocks? Why to copy&paste if you can externalise   it?

pkg/pquerier/collector.go
talIguaz: that would be a one liner with 3 parameters.. not sure if it would simplify the code

pkg/pquerier/collector.go
talIguaz: then i would need to also cast `currentCell` to int anyway so it doesn't save a cast

pkg/pquerier/collector.go
talIguaz: Not exactly, the computation of `currentCell` is different. 
In the `intervalAggregation` i look at the "step" **after** the current point, and in the `windowAggregation` i look at the "step" **before** the current time.
I guess i can create another shared function but for logical separation i would prefer to leave it as is.

pkg/pquerier/query_integration_test.go
talIguaz: take a look at https://github.com/v3io/v3io-tsdb/pull/326 😄

pkg/pquerier/query_integration_test.go
talIguaz: every test has different input data and expected data, so the overhead of trying to move it all to a generic "setup" function wouldn't be that "clean" i think. but worth thinking about it in another PR 😄

pkg/pquerier/collector.go
imakhlin: it will not simplify but will localise the possible changes in the logic and stop duplicating same code

pkg/pquerier/collector.go
imakhlin: just add before/after argument to the function

pkg/pquerier/query_integration_test.go
imakhlin: break it into smaller logical units/files

pkg/pquerier/query_integration_test.go
imakhlin: ok, remember that big brother is watching...

PR #325, opened 2019-03-14
Explicitly request client aggr

talIguaz:
add an option for the user to explicitly request client side aggregations.
Pros: more accurate than server side
Cons: might worsen performance

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -164,6 +164,8 @@ type V3ioConfig struct {
 	BuildInfo *BuildInfo `json:"buildInfo,omitempty"`
 	// Override nginx bug
 	DisableNginxMitigation bool `json:"disableNginxMitigation,omitempty"`
+	// explicitly always use client aggregation
+	UsePreciseAggregations bool `json:"usePreciseAggregations,omitempty"`
 }
 
 type MetricsReporterConfig struct {
--- a/pkg/pquerier/query_integration_test.go
+++ b/pkg/pquerier/query_integration_test.go
@@ -3286,6 +3286,63 @@ func (suite *testQuerySuite) TestSelectDataframeDaownsampleMetricsHaveBigGaps() 
 	suite.Require().Equal(1, dataFrameCount, "series count didn't match expected")
 }
 
+func (suite *testQuerySuite) TestUsePreciseAggregationsConfig() {
+	suite.v3ioConfig.UsePreciseAggregations = true
+	defer func() { suite.v3ioConfig.UsePreciseAggregations = false }()
+	adapter, err := tsdb.NewV3ioAdapter(suite.v3ioConfig, nil, nil)
+	if err != nil {
+		suite.T().Fatalf("failed to create v3io adapter. reason: %s", err)
+	}
+	labels1 := utils.LabelsFromStringList("os", "linux")
+	numberOfEvents := 10
+	eventsInterval := 60 * 1000
+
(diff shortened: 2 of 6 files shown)


pkg/pquerier/query_integration_test.go
```diff
@@ -3286,6 +3286,63 @@ func (suite *testQuerySuite) TestSelectDataframeDaownsampleMetricsHaveBigGaps()
...
+		data, err := tsdbtest.IteratorToSlice(iter)
+		agg := set.At().Labels().Get(aggregate.AggregateLabel)
+		if err != nil {
+			suite.T().Fatal(err)
+		}
+
+		assert.Equal(suite.T(), expected[agg], data, "queried data does not match expected")
+	}
+
+	assert.Equal(suite.T(), 3, seriesCount, "series count didn't match expected")
```
gtopper: `suite.Require()`?

pkg/config/config.go
```diff
@@ -164,6 +164,8 @@ type V3ioConfig struct {
 	BuildInfo *BuildInfo `json:"buildInfo,omitempty"`
 	// Override nginx bug
 	DisableNginxMitigation bool `json:"disableNginxMitigation,omitempty"`
+	// explicitly always use client aggregation
+	UsePreciseAggregations bool `json:"usePreciseAggregations,omitempty"`
```
gtopper: So why not call it `UseClientAggregations`?

pkg/tsdbctl/query.go
```diff
@@ -113,6 +114,8 @@ Arguments:
 		"Aggregation interval for applying the aggregation functions\n(if set - see the -a|--aggregates flag), of the format\n\"[0-9]+[mhd]\" (where 'm' = minutes, 'h' = hours, and\n'd' = days). Examples: \"1h\"; \"150m\". (default =\n<end time> - <start time>)")
 	cmd.Flags().StringVar(&commandeer.groupBy, "groupBy", "",
 		"Comma separated list of labels to group the result by")
+	cmd.Flags().BoolVar(&commandeer.usePreciseAggregations, "use-precise-aggregations", false,
+		"Ignore server aggregations optimizations for more accurate results.")
```
gtopper: `Ignore server aggregations optimizations` is not correct English :-). You can change `aggregations` -> `aggregation`, drop the `optimizations `, or rephrase.

pkg/pquerier/query_integration_test.go
```diff
@@ -3286,6 +3286,63 @@ func (suite *testQuerySuite) TestSelectDataframeDaownsampleMetricsHaveBigGaps()
 	suite.Require().Equal(1, dataFrameCount, "series count didn't match expected")
 }
 
+func (suite *testQuerySuite) TestUsePreciseAggregationsConfig() {
+	suite.v3ioConfig.UsePreciseAggregations = true
+	defer func() { suite.v3ioConfig.UsePreciseAggregations = false }()
+	adapter, err := tsdb.NewV3ioAdapter(suite.v3ioConfig, nil, nil)
+	if err != nil {
+		suite.T().Fatalf("failed to create v3io adapter. reason: %s", err)
```
gtopper: Use `NoError`?

pkg/config/config.go
talIguaz: thought this would be clearer for the user. The user dont care/realy know about server vs client aggregates but he does know what a Precise aggregation is. 
WDYT?

gshatz: I think we can use this name and Ask @Sharon-iguazio to document the implications

PR #326, opened 2019-03-14
Refactor querier tests

talIguaz:
Group tests into suites based on features


imakhlin:
break it into smaller files

talIguaz:
Not sure i want a more fine-grained split. already grouped enough logically..
Also tests files are "supposed" to be longer than normal code files, and when you go through the tests it is usually enough to just read the test names (or at least should be enough)

PR #337, opened 2019-03-26
Add server aggregation coefficient config

talIguaz:
* move coefficient between query step and aggregation granularity, which determines whether or not to use server side aggregations to a configurable parameter.
* dont use server aggregates when the aggr granularity is divider of the Step. *Only* by the coefficient mentioned above
* fix downsample bug that returned incorrect results (found out while refactoring tests)


pkg/pquerier/collector.go
```diff
@@ -196,24 +196,29 @@ func downsampleRawData(ctx *selectQueryContext, res *qryResults,
...
-					prevV = previousPartitionLastValue
-				}
-				interpolatedT, interpolatedV := col.GetInterpolationFunction()(prevT, t, currCellTime, prevV, v)
+		prev, err := col.getBuilder().At(currCell)
+
+		// Only update a cell if it hasn't been set yet
+		if prev == nil || err != nil {
+			if it.Seek(currCellTime) {
+				t, v := it.At()
+				//tCellIndex := (t - ctx.queryParams.From) / ctx.queryParams.Step
```
gtopper: 😮

pkg/aggregate/aggregationParams.go
```diff
@@ -80,3 +83,12 @@ func (as *AggregationParams) GetAttrNames() []string {
 
 	return names
 }
+
+func (as *AggregationParams) getServerAggregateCoefficient() int64 {
+	conf, err := config.GetOrDefaultConfig()
+	if err != nil {
+		return config.DefaultUseServerAggregateCoefficient
```
gtopper: Don't we generally put the default in the config or commandeer?

pkg/aggregate/aggregationParams.go
```diff
@@ -24,13 +27,14 @@ func NewAggregationParams(functions, col string, buckets int, interval, aggregat
...
-		Interval:          interval,
-		overlapWindows:    windows,
+		aggrMask:                 aggrMask,
+		colName:                  col,
+		buckets:                  buckets,
+		rollupTime:               rollupTime,
+		aggregationWindow:        aggregationWindow,
+		Interval:                 interval,
+		overlapWindows:           windows,
+		disableClientAggregation: disableClientAggregate,
```
gtopper: Please make up your mind what to call it 😄

pkg/aggregate/aggregationParams.go
```diff
@@ -44,8 +48,7 @@ func (as *AggregationParams) CanAggregate(partitionAggr AggrType) bool {
 	// if interval and rollup are not even divisors we need higher resolution (3x) to smooth the graph
 	// when we add linear/spline graph projection we can reduce back to 1x
 	return ((aggrMask & partitionAggr) == aggrMask) &&
-		as.Interval >= as.rollupTime &&
-		(as.Interval%as.rollupTime == 0 || as.Interval/as.rollupTime > 3) &&
+		(as.Interval/as.rollupTime > as.getServerAggregateCoefficient() || (as.Interval == as.rollupTime && as.disableClientAggregation)) &&
```
gtopper: If `Interval` is `rollupTime`*2, we don't want to use server aggregations anymore? Also not sure I understand what `disableClientAggregation` is for.

PR #343, opened 2019-04-03
TSDB-91

talIguaz:
Fixed 2 bugs:
* When requesting `disableClientAggregate=true` and we dont have the server aggregates for the query an empty result was returned.
* When doing 1 query for 2 metrics with different aggregates where 1 is in the server and 1 not, an exception occurred.

### DIFF
--- a/pkg/pquerier/select.go
+++ b/pkg/pquerier/select.go
@@ -219,11 +219,17 @@ func (queryCtx *selectQueryContext) queryPartition(partition *partmgr.DBPartitio
 		for _, q := range queries {
 			q.aggregationParams = nil
 			q.useServerSideAggregates = false
-			q.getItems(queryCtx)
+			err = q.getItems(queryCtx)
+			if err != nil {
+				break
+			}
 		}
 	} else {
 		for _, q := range queries {
-			q.getItems(queryCtx)
+			err = q.getItems(queryCtx)
+			if err != nil {
+				break
+			}
 		}
 	}
 
--- a/pkg/pquerier/select.go
+++ b/pkg/pquerier/select.go
@@ -36,6 +36,10 @@ type selectQueryContext struct {
 	totalColumns           int
 	isCrossSeriesAggregate bool
 
+	// In case one of the aggregates of one of the metrics should use client side aggregates
+	// but the user requested to disable client aggregations - return raw data for every requested metric
+	forceRawQuery bool
+
 	dataFrames      map[uint64]*dataFrame
 	frameList       []*dataFrame
 	requestChannels []chan *qryResults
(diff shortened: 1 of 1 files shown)


pkg/pquerier/select.go
```diff
@@ -183,12 +195,36 @@ func (queryCtx *selectQueryContext) queryPartition(partition *partmgr.DBPartitio
...
+		}
+		index++
+	}
+
+	if doForceAllRawQuery {
+		queryCtx.forceRawQuery = true
+		for _, q := range queries {
+			q.aggregationParams = nil
+			q.useServerSideAggregates = false
+			q.getItems(queryCtx)
```
gtopper: What about err?

pkg/pquerier/select.go
```diff
@@ -183,12 +195,36 @@ func (queryCtx *selectQueryContext) queryPartition(partition *partmgr.DBPartitio
...
+		}
+	} else {
+		for _, q := range queries {
+			q.getItems(queryCtx)
```
gtopper: What about err?

pkg/pquerier/select.go
```diff
@@ -183,12 +195,42 @@ func (queryCtx *selectQueryContext) queryPartition(partition *partmgr.DBPartitio
 			}
 		}
 
-		var preAggregateLabels []string
 		if newQuery.useServerSideAggregates && !requestAggregatesAndRaw {
-			preAggregateLabels = queryCtx.parsePreAggregateLabels(partition)
+			newQuery.preAggregateLabels = queryCtx.parsePreAggregateLabels(partition)
+		} else {
+			newQuery.preAggregateLabels = []string{}
```
gtopper: Why do this?

pkg/pquerier/select.go
```diff
@@ -183,12 +195,40 @@ func (queryCtx *selectQueryContext) queryPartition(partition *partmgr.DBPartitio
...
+	}
+
+	if doForceAllRawQuery {
+		queryCtx.forceRawQuery = true
+		for _, q := range queries {
+			q.aggregationParams = nil
+			q.useServerSideAggregates = false
+			err = q.getItems(queryCtx)
+			if err != nil {
+				break
```
gtopper: Sure... err is always `nil` 🙈

pkg/pquerier/select.go
```diff
@@ -144,6 +148,9 @@ func (queryCtx *selectQueryContext) queryPartition(partition *partmgr.DBPartitio
 		mint = queryCtx.queryParams.From
 	}
 
+	queryRawInsteadOfAggregates, doForceAllRawQuery := false, false
```
gtopper: Could be `var queryRawInsteadOfAggregates, doForceAllRawQuery bool` 🙂

PR #366, opened 2019-04-16
Fix command line server parameter

gtopper:
dropping backward compatibility with older formats.

### DIFF
--- a/pkg/tsdbctl/tsdbctl_test.go
+++ b/pkg/tsdbctl/tsdbctl_test.go
@@ -37,8 +37,8 @@ type testTsdbctlSuite struct {
 }
 
 func (suite *testTsdbctlSuite) TestPopulateConfigWithTenant() {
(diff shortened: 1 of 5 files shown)


pkg/tsdbctl/tsdbctl_test.go
```diff
@@ -52,11 +52,11 @@ func (suite *testTsdbctlSuite) TestPopulateConfigWithTenant() {
 
 	expectedRc := RootCommandeer{
 		v3iocfg:  cfg,
-		v3ioPath: "localhost:80123/123",
+		v3ioUrl:  "localhost:80123",
```
talIguaz: how is the test passes if you deleted the `container` from the expected and not from the "input"

gtopper: Because it goes in the `container` field, not `v3ioUrl`.

PR #371, opened 2019-05-07
TSDB 112 + TSDB-113

talIguaz:
Add SQL parsing error checks:
* cannot use same alias multiple times (tsdb-113)
* cannot use alias on wildcard
* cannot use metric name as alias (tsdb-112)

### DIFF
--- a/pkg/pquerier/sql_parser.go
+++ b/pkg/pquerier/sql_parser.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/v3io/v3io-tsdb/pkg/utils"
 	"github.com/xwb1989/sqlparser"
 )
@@ -119,7 +120,7 @@ func parseFuncExpr(expr *sqlparser.FuncExpr, destCol *RequestedColumn) error {
 		}
 
 		if destCol.Metric == "" && destCol.Alias != "" {
-			return fmt.Errorf("cannot alias a wildcard")
+			return errors.New("cannot alias a wildcard")
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/pquerier/sql_parser.go
```diff
@@ -109,6 +117,10 @@ func parseFuncExpr(expr *sqlparser.FuncExpr, destCol *RequestedColumn) error {
 				parseFuncExpr(innerExpr, destCol)
 			}
 		}
+
+		if destCol.Metric == "" && destCol.Alias != "" {
+			return fmt.Errorf("cannot alias a wildcard")
```
gtopper: Better use `errors.New`.

pkg/pquerier/sql_parser.go
```diff
@@ -139,3 +151,25 @@ func parseFilter(originalFilter string) (string, error) {
...
 }
+
+func validateColumnNames(params *SelectParams) error {
+	names := make(map[string]bool)
+	requestedMetrics := make(map[string]bool)
+
+	for _, column := range params.RequestedColumns {
+		columnName := column.GetColumnName()
+		if names[columnName] {
+			return fmt.Errorf("column name '%v' apears more than once in select query", columnName)
```
gtopper: `apears` -> `appears`

pkg/pquerier/sql_parser_test.go
```diff
@@ -76,3 +76,22 @@ func TestParseQuery(t *testing.T) {
...
+	}{
+		{input: "select columnA as something, columnB as something"},
+		{input: "select avg(columnA) as something, columnB as something"},
+		{input: "select avg(*) as something"},
+		{input: "select avg(cpu), max(cpu) as cpu"},
+	}
+	for _, test := range testCases {
+		t.Run(test.input, func(tt *testing.T) {
+			_, _, err := pquerier.ParseQuery(test.input)
+			if err == nil {
```
gtopper: Maybe be more specific?

talIguaz: there isn't much more i can output, since there is not error.
I can add something like an "expected error" to the test case struct but, mehh.. I think its clear enough and you'll have to debug it anyway.

PR #376, opened 2019-06-06
Add feature flag to load partitions from schema file

gtopper:
as opposed to the new method of loading them from KV.


gtopper:
Will start a new CI system on Monday...

talIguaz:
i think we need to also check this config on write, so that we won't bump schema version to 3.

PR #380, opened 2019-06-12
Use go mod vendor for tsdb-nuclio.

gtopper:


gshatz:
@gtopper please do the same for tsdb-query

PR #396, opened 2019-08-12
IG-12226

talIguaz:
Output of the check command in string values should be the same as regular metrics.
Example:
```
	Attr: _v13
		Unix timestamp=1565536693000, t=2019-08-11T15:18:13Z, v=dildo
```


pkg/tsdbctl/check.go
```diff
@@ -311,3 +324,12 @@ func getSchema(cfg *config.V3ioConfig, container v3io.Container) (*config.Schema
 	}
 	return tableSchema, nil
 }
+
+func getEncoding(enc string) (chunkenc.Encoding, error) {
+	intEncoding, err := strconv.Atoi(enc)
+	if err != nil {
+		return 0, fmt.Errorf("error parsing encoding type of chunk, got: %v, error: %v", enc, err)
```
imakhlin: tell that encoding should have a numeric value

PR #397, opened 2019-08-13
IG-12645

talIguaz:
return columns of **raw** query by requested order


pkg/pquerier/frames.go
```diff
@@ -106,6 +106,23 @@ func NewDataFrame(columnsSpec []columnMeta, indexColumn Column, lset utils.Label
 	// is raw query
 	if isRawQuery {
 		df.columnByName = make(map[string]int, len(columnsSpec))
+		df.rawColumns = make([]utils.Series, len(columnsSpec))
```
gtopper: Can init this at the end of this new snippet if needed.

PR #413, opened 2019-12-31
IG 13865

talIguaz:


pkg/pquerier/frames.go
```diff
@@ -469,16 +469,20 @@ func (d *dataFrame) rawSeriesToColumns() error {
 	nextTime := int64(math.MaxInt64)
 	seriesHasMoreData := make([]bool, len(d.rawColumns))
 
+	emptyMetrics := make(map[int]string)
+
 	for i, rawSeries := range d.rawColumns {
 		if rawSeries == nil {
 			missingColumn := "(unknown column)"
 			for columnName, index := range d.columnByName {
 				if index == i {
```
gtopper: Not new in this PR, but it looks like this could basically just be a map lookup instead of a for loop?

pkg/pquerier/frames.go
```diff
@@ -550,13 +557,34 @@ func (d *dataFrame) rawSeriesToColumns() error {
...
+		nullValues := make([]float64, numberOfRows)
+		for i := 0; i < numberOfRows; i++ {
+			nullValues[i] = math.NaN()
+		}
+		for index, metricName := range emptyMetrics {
+			spec := columnMeta{metric: metricName}
+			col := NewDataColumn(metricName, spec, numberOfRows, frames.FloatType)
+			framesCol, err := frames.NewSliceColumn(metricName, nullValues)
+			if err != nil {
+				return err
```
gtopper: Probably good to wrap this one.

pkg/pquerier/querier.go
```diff
@@ -183,6 +183,14 @@ func (q *V3ioQuerier) baseSelectQry(params *SelectParams, showAggregateLabel boo
 			return
 		}
 
+		var existingMetrics map[string]bool
+		existingMetrics, err = q.getExistingMetricNamesMap()
+
+		if err != nil {
```
gtopper: Strange vertical spacing before checking `err`.

pkg/pquerier/select.go
```diff
@@ -374,6 +376,10 @@ func (queryCtx *selectQueryContext) createColumnSpecs() ([]columnMeta, map[strin
 	}
 
 	for i, col := range requestedColumns {
+		if col.Metric != "" && !queryCtx.existingMetrics[col.Metric] {
```
gtopper: Not critical, but `existingMetrics` would be clearer as an argument rather than receiver state.

pkg/pquerier/frames.go
talIguaz: Not really, cause you dont have the column name (the map key), you are trying to find it based on the index.

PR #417, opened 2020-01-02
IG-13690: Filter out nil raw series on read.

gtopper:


pkg/pquerier/frames.go
```diff
@@ -56,22 +56,27 @@ func (fi *frameIterator) Next() bool {
...
+		fi.setIndex++
+		fi.seriesIndex = 0
 	}
 
-	fi.setIndex++
-	fi.seriesIndex = 0
 	if fi.isCurrentSeriesHidden() {
 		return fi.Next()
 	}
+	if len(fi.ctx.frameList) > 0 {
```
talIguaz: looks like a redundant check, `len(fi.ctx.frameList)` can't be 0 at this point, right?

pkg/pquerier/frames.go
```diff
@@ -56,25 +56,22 @@ func (fi *frameIterator) Next() bool {
...
 		fi.seriesIndex++
-	} else {
+	} else if fi.setIndex+1 >= len(fi.ctx.frameList) {
 		// already in the last column in the last frame
-		if fi.setIndex+1 >= len(fi.ctx.frameList) {
-			return false
-		}
-
+		return false
+	} else {
```
talIguaz: If there are already comments to every case, maybe add also comment for the last `else`

PR #418, opened 2020-01-05
IG-14353

talIguaz:


dinal:
do we have jupy notebooks that use next/prev?

PR #420, opened 2020-01-09
IG-14485: Avoid crash on empty data in cross series query.

gtopper:


talIguaz:
Add a test 
#420

gtopper:
I tried, but ended up running into other bugs with cross series aggregations instead... I'll check with Sahar if he can provide a reproduce procedure, to avoid spending too much time on this.

PR #453, opened 2020-02-26
IG 14978

talIguaz:
Start new xor series if time delta is larger then 32bit int

### DIFF
--- a/pkg/chunkenc/xor.go
+++ b/pkg/chunkenc/xor.go
@@ -160,6 +160,8 @@ type xorAppender struct {
 
 	leading  uint8
 	trailing uint8
+
+	isPreviousNewSeries bool
 }
 
 func (a *xorAppender) Encoding() Encoding {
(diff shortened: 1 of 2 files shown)


pkg/chunkenc/xor.go
```diff
@@ -186,37 +188,44 @@ func (a *xorAppender) Append(t int64, vvar interface{}) {
...
-
-	} else if num == 1 {
+		a.isPreviousNewSeries = true
+	} else {
 		tDelta = uint64(t - a.t)
 
-		a.b.writeBits(tDelta, 32)
-		a.writeVDelta(v)
+		// We write time deltas as 32 bits (for compression) if the delta is too large we'll start a new series
+		if bits.Len64(tDelta) >= 32 {
```
imakhlin: you can combine `num == 0` and `bits.Len64(tDelta) >= 32` into `a.hasToStartNewSeries(t)` and then reuse the code of initial block

dinal:
pls open the pr against dev, it's not going to 2.5 after all

PR #457, opened 2020-02-26
IG-15002

talIguaz:


pkg/tsdb/v3iotsdb.go
```diff
@@ -353,7 +353,7 @@ func (a *V3ioAdapter) DeletePartitionsData(deleteParams *DeleteParams) error {
 		}
 	}
 
-	for i := 0; i <= a.cfg.Workers; i++ {
+	for i := 0; i < a.cfg.Workers; i++ {
```
imakhlin: I personally don't like counting approach. I prefer using context or waiting groups

PR #471, opened 2020-03-18
IG-14149

talIguaz:
- fix sporadic integration test
- fix error handling in collector, and stop go-routines on errors
- read query responses concurrently rather then sequential


go.mod
```diff
@@ -13,9 +14,11 @@ require (
 	github.com/pkg/errors v0.8.1
 	github.com/rcrowley/go-metrics v0.0.0-20181016184325-3113b8401b8a
 	github.com/spf13/cobra v0.0.3
+	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/stretchr/testify v1.4.0
 	github.com/v3io/frames v0.6.14-v0.9.18
 	github.com/v3io/v3io-go v0.1.5-0.20200316155009-cc83de374ff2
+	github.com/v3io/v3io-go-http v0.0.0-20190415143924-cc2fbcde6663 // indirect
```
gtopper: Eh?

talIguaz: `go mod` 🤷‍♂

gtopper: We have to figure out where it came from and fix it. v3io-go-http is not supposed to be used anymore, definitely not in parallel with v3io-go.

pkg/pquerier/select.go
```diff
@@ -349,19 +376,30 @@ func (queryCtx *selectQueryContext) processQueryResults(query *partQuery) error
 				results.IsServerAggregates(),
 				queryCtx.showAggregateLabel)
 			if err != nil {
-				return err
+				queryCtx.errorChannel <- err
+				queryCtx.createDFLock.Unlock()
```
gtopper: Are you sure writing to `queryCtx.errorChannel <- err` can't block, potentially deadlocking `createDFLock`?

go.mod
gtopper: https://github.com/v3io/v3io-tsdb/pull/472

PR #476, opened 2020-03-24
IG-15260

talIguaz:
query 2 last partitions instead of only first one

### DIFF
--- a/pkg/pquerier/querier.go
+++ b/pkg/pquerier/querier.go
@@ -321,12 +321,16 @@ func (q *V3ioQuerier) GetLabelSets(metric string, filter string) ([]utils.Labels
 
 	// Get all label sets
 	input := v3io.GetItemsInput{
-		Path:           partitionPaths[0],
 		Filter:         filter,
 		AttributeNames: []string{config.LabelSetAttrName, config.MetricNameAttrName},
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/pquerier/querier.go
```diff
@@ -321,12 +321,16 @@ func (q *V3ioQuerier) GetLabelSets(metric string, filter string) ([]utils.Labels
 
 	// Get all label sets
 	input := v3io.GetItemsInput{
-		Path:           partitionPaths[0],
 		Filter:         filter,
 		AttributeNames: []string{config.LabelSetAttrName, config.MetricNameAttrName},
 	}
 
-	iter, err := utils.NewAsyncItemsCursor(q.container, &input, q.cfg.QryWorkers, shardingKeys, q.logger)
+	// because of performance issues we only want to query the first and last partitions
```
dinal: comment is incorrect

pkg/pquerier/querier.go
```diff
@@ -321,12 +321,16 @@ func (q *V3ioQuerier) GetLabelSets(metric string, filter string) ([]utils.Labels
...
+	partitionsToQuery := []string{partitionPaths[len(partitionPaths)-1]}
```
dinal: can partitions be empty? then it's size is 0

talIguaz: https://github.com/v3io/v3io-tsdb/pull/476/files#diff-44f0b6ed4756e6b5b32c5a5b19a6d5d4R311

PR #481, opened 2020-03-26
IG-15233 fix hang in ingest

dinal:
there are 2 issues that cause hang in ingest - 
1. when all requests were sent and answered and the only remaining item in asyncChannel is the waitForcomplition req it didn't enter again to the loop to see that everything is done and mark 'gotCompletion=true'
2. when all requests that were sent are answered and the remaining samples to ingest are out of the partition range, we push them again to metricsQueue but nothing makes it enter the loop again to read those metrics (we pop only on received updates)

### DIFF
--- a/pkg/appender/ingest.go
+++ b/pkg/appender/ingest.go
@@ -50,6 +50,7 @@ func (mc *MetricsCache) metricFeed(index int) {
 		inFlight := 0
 		gotData := false
 		gotCompletion := false
+		potentialCompletion := false
 		var completeChan chan int
 
 		for {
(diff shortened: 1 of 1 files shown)


pkg/appender/ingest.go
```diff
@@ -250,8 +258,14 @@ func (mc *MetricsCache) postMetricUpdates(metric *MetricState) {
 		} else if sent {
 			metric.setState(storeStateUpdate)
 		}
-		if !sent && metric.store.samplesQueueLength() == 0 {
-			metric.setState(storeStateReady)
+		if !sent {
```
talIguaz: we need to also do some handling for the other `writeChunks` call, no?
in `ingest.go:353`

pkg/appender/ingest.go
```diff
@@ -78,12 +84,14 @@ func (mc *MetricsCache) metricFeed(index int) {
 					if app.metric == nil {
 						// Handle update completion requests (metric == nil)
 						completeChan = app.resp
-
 						length := mc.metricQueue.Length()
-						if gotCompletion && length == 0 && len(mc.asyncAppendChan) == 0 {
-							completeChan <- 0
-							gotCompletion = false
-							gotData = false
+						if length == 0 && len(mc.asyncAppendChan) == 0 {
```
talIguaz: aren't you suppose to set `potentialCompletion` to `false` in the else of this condition?

dinal: yep it needs to be set to false but it should be false in any case after we read next item from asyncChannel
if it's waitForComplition req then we are done and if not then set to false and move on

PR #489, opened 2020-04-01
IG-15372: Avoid failing delete on 404.

gtopper:
Avoid trying to delete partition dir if an error occurred when deleting its content.

### DIFF
--- a/pkg/tsdb/v3iotsdb.go
+++ b/pkg/tsdb/v3iotsdb.go
@@ -476,7 +476,7 @@ func deleteEntirePartition(logger logger.Logger, container v3io.Container, parti
 	defer wg.Done()
 
 	err := utils.DeleteTable(logger, container, partitionPath, "", workers)
-	if err != nil && !utils.IsNotExistsError(err) {
+	if err != nil {
 		errChannel <- errors.Wrapf(err, "Failed to delete partition '%s'.", partitionPath)
 		return
 	}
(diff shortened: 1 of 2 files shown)


pkg/utils/container.go
```diff
@@ -124,18 +123,17 @@ func DeleteTable(logger logger.Logger, container v3io.Container, path, filter st
...
-	reqMap := map[uint64]bool{}
+	doneChan := make(chan error, 1)
+	go respWaitLoop(commChan, responseChan, doneChan, 10*time.Second)
 
 	i := 0
 	for iter.Next() {
 		name := iter.GetField(config.ObjectNameAttrName).(string)
-		req, err := container.DeleteObject(&v3io.DeleteObjectInput{Path: path + "/" + name}, nil, responseChan)
+		_, err := container.DeleteObject(&v3io.DeleteObjectInput{Path: path + "/" + name}, nil, responseChan)
 		if err != nil {
```
talIguaz, line 133: Need to check the status code here, so that the worker won't quit and we will continue deleting all remaining items

pkg/tsdb/v3iotsdb.go
```diff
@@ -476,12 +476,13 @@ func deleteEntirePartition(logger logger.Logger, container v3io.Container, parti
 	defer wg.Done()
 
 	err := utils.DeleteTable(logger, container, partitionPath, "", workers)
-	if err != nil {
+	if err != nil && !utils.IsNotExistsError(err) {
```
talIguaz, line 479: move this check to my other comment 😄

pkg/utils/container.go
gtopper: This error is only for sending the request. The response is received asynchronously.

pkg/tsdb/v3iotsdb.go
gtopper: I would, except it would break the code :)

PR #495, opened 2020-04-30
[IG-15609] handle error in NextPart

dinal:


pkg/appender/store.go
```diff
@@ -400,7 +410,7 @@ func (cs *chunkStore) writeChunks(mc *MetricsCache, metric *MetricState) (hasPen
 			cs.pending = cs.pending[pendingSampleIndex:]
 		}
 
-		if pendingSamplesCount == 0 || expr == "" {
+		if pendingSamplesCount == 0 || expr == "" || err != nil {
```
talIguaz, line 413: why do we need to check the error again here?

pkg/appender/store.go
```diff
@@ -255,29 +255,34 @@ func (cs *chunkStore) chunkByTime(t int64, isVariantEncoding bool) *attrAppender
 		chunk := chunkenc.NewChunk(cs.logger, isVariantEncoding) // TODO: init based on schema, use init function
 		app, err := chunk.Appender()
 		if err != nil {
-			return nil
+			cs.logger.Error(err)
+			return nil, err
+		}
+		nextPart, err := part.NextPart(t)
+		if err != nil {
+			cs.logger.Error(err)
```
talIguaz, line 263: no need to log i guess, it will be returned to the user, no?

pkg/appender/store.go
```diff
@@ -255,29 +255,34 @@ func (cs *chunkStore) chunkByTime(t int64, isVariantEncoding bool) *attrAppender
 		chunk := chunkenc.NewChunk(cs.logger, isVariantEncoding) // TODO: init based on schema, use init function
 		app, err := chunk.Appender()
 		if err != nil {
-			return nil
+			cs.logger.Error(err)
```
talIguaz, line 258: you are returning here an err (where previously it wasn't returned), which might affect the flow in the condition at line 350.
so better understand why the error was ignored before

pkg/appender/store.go
dinal: yes, cause if we existed the loop with an error i want to push the metric back to queue

pkg/appender/store.go
dinal: it's הכנה למזגן, there's no Appender that actually returns an error

PR #506, opened 2020-07-16
IG-16343: Ingest loop fixes to avoid data loss and deadlocks.

gtopper:


dinal:
double check these changes with a few stress tool runs (before adrian's tweaking)
/opt/tools/framesd/framesd --serv-addr=https://framesd.default-tenant.app.dev71.lab.iguazeng.com --backend=tsdb --operation=write --workers-count=30 --run-time=240 --records-count=1000 --container=bigdata --cleanup --table-max-records=100000

PR #515, opened 2020-08-04
Fix linting errors.

gtopper:


pkg/tsdb/v3iotsdb.go
```diff
@@ -202,12 +201,6 @@ func (a *V3ioAdapter) connect() error {
 		return errors.Wrapf(err, "Failed to unmarshal the TSDB schema at '%s', got: %v .", fullpath, string(resp.Body()))
 	}
 
-	// in order to support backward compatibility we do not fail on version mismatch and only logging warning
-	if a.cfg.LoadPartitionsFromSchemaAttr && tableSchema.TableSchemaInfo.Version != schema.Version {
```
talIguaz, line 206: Maybe keep the version mismatch warning, might help us understand potential bugs in the future.

gtopper: Maybe? I'm not sure why this would be helpful...

talIguaz (approved):

PR #523, opened 2020-08-17
IG-16619 

talIguaz:
Get state when reaching new partition

### DIFF
--- a/pkg/appender/ingest.go
+++ b/pkg/appender/ingest.go
@@ -65,9 +65,7 @@ func (mc *MetricsCache) metricFeed(index int) {
 					// Handle completion notifications from the update loop
 					if app.isCompletion {
 						gotCompletion = true
-						continue
-					}
-					if app.metric == nil {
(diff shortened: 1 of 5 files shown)


pkg/appender/ingest.go
```diff
@@ -223,45 +228,70 @@ func (mc *MetricsCache) postMetricUpdates(metric *MetricState) {
...
-			// Count errors
-			mc.performanceReporter.IncrementCounter("GetChunksStateError", 1)
-
-			mc.logger.ErrorWith("Failed to get item state", "metric", metric.Lset, "err", err)
-			setError(mc, metric, err)
-		} else {
-			metric.setState(storeStateGet)
+	// In case we are in pre get state or our data spreads across multiple partitions, get the new state for the current partition
+	if metric.getState() == storeStatePreGet ||
+		(metric.canSendRequests() && metric.shouldGetState) {
```
gtopper, line 234: No need to break. No need for parens.

PR #533, opened 2020-10-07
IG-15394 change metric cache to be lru cache

dinal:


pkg/appender/appender.go
```diff
@@ -147,8 +151,7 @@ func NewMetricsCache(container v3io.Container, logger logger.Logger, cfg *config
 	partMngr *partmgr.PartitionManager) *MetricsCache {
 
 	newCache := MetricsCache{container: container, logger: logger, cfg: cfg, partitionMngr: partMngr}
-	newCache.cacheMetricMap = map[cacheKey]*MetricState{}
-	newCache.cacheRefMap = map[uint64]*MetricState{}
+	newCache.cacheMetricMap = lru.New(cfg.MetricCacheSize) //map[cacheKey]*MetricState{}
```
gtopper, line 154: Dead code 😵

pkg/appender/appender.go
```diff
@@ -277,29 +273,55 @@ func (mc *MetricsCache) Add(lset utils.LabelsIfc, t int64, v interface{}) (uint6
...
 		mc.appendTV(aggrMetric, t, v)
 	}
 
-	return metric.refID, err
+	return &MetricIdentifier{metric.name, metric.hash}, err
 }
 
-// fast Add to metric (by refID)
-func (mc *MetricsCache) AddFast(ref uint64, t int64, v interface{}) error {
+func (mc *MetricsCache) AddFast(lset utils.Labels, ref *MetricIdentifier, t int64, v interface{}) error {
```
gtopper, line 287: * `lset` is not used
* Maybe rename `ref` now that it's a `MetricIdentifier`?

pkg/appender/appender.go
```diff
@@ -277,29 +273,55 @@ func (mc *MetricsCache) Add(lset utils.LabelsIfc, t int64, v interface{}) (uint6
...
 
 	err := verifyTimeValid(t)
 	if err != nil {
 		return err
 	}
+	if ref == nil {
+		return fmt.Errorf("nil identifier")
```
gtopper, line 294: You can avoid this by making the parameter `MetricIdentifier` rather than `*MetricIdentifier`. I promise you that using a pointer isn't helping performance in this case 🙂

pkg/appender/appender.go
```diff
@@ -277,29 +273,55 @@ func (mc *MetricsCache) Add(lset utils.LabelsIfc, t int64, v interface{}) (uint6
...
+	}
+	metric, ok := mc.getMetric(ref.name, ref.hash)
+	if !ok {
+		mc.logger.ErrorWith("metric not found", "name", ref.name, "hash", ref.hash)
+		return fmt.Errorf("metric not found")
```
gtopper, line 299: Probably no need to log an error if we return an error.

pkg/appender/appender.go
```diff
@@ -277,29 +273,55 @@ func (mc *MetricsCache) Add(lset utils.LabelsIfc, t int64, v interface{}) (uint6
...
+	mc.appendTV(metric, t, v)
+
+	for _, aggrMetric := range metric.aggrs {
+		mc.appendTV(aggrMetric, t, v)
+	}
+
+	return err
+}
+
+func (mc *MetricsCache) AddFastWithMetric(lset utils.Labels, t int64, v interface{}) error {
```
gtopper, line 314: Not used. According to @dinal, this was added accidentally.

test/integration/aggregates/aggregates_integration_test.go
```diff
@@ -24,6 +24,7 @@ package aggregates
 
 import (
 	"fmt"
+	"github.com/v3io/v3io-tsdb/pkg/appender"
```
gtopper, line 27: Usually impi freaks out about this.

pkg/appender/appender.go
```diff
@@ -277,29 +273,55 @@ func (mc *MetricsCache) Add(lset utils.LabelsIfc, t int64, v interface{}) (uint6
...
+	if err != nil {
+		return err
+	}
+	name, _, hash := lset.GetKey()
+	metric, ok := mc.getMetric(name, hash)
 	if !ok {
-		mc.logger.ErrorWith("Ref not found", "ref", ref)
-		return fmt.Errorf("ref not found")
+		mc.logger.ErrorWith("metric not found", "name", name, "hash", hash)
+		return fmt.Errorf("metric not found")
```
gtopper, line 324: Might want to remove the log here as well.

test/integration/aggregates/aggregates_integration_test.go
dinal: we don't run impi on tests but i'll move it anyway

pkg/appender/appender.go
dinal: i prefer a pointer so i could return nil on errors instead of empty MetricIdentifier, looks nicer
any reason not to use a pointer?

gtopper: Ah, I see. But here it's a parameter, not a return value :)
Only reason here is to avoid a bad argument to begin with.

test/integration/aggregates/aggregates_integration_test.go
gtopper: I thought so, but in the makefile it looks like it runs on `./...`. Oh well.

gtopper: Not a fan of impi anyway.

PR #535, opened 2020-10-08
use hash with name as key in metric cache

dinal:


pkg/utils/labels.go
```diff
@@ -69,7 +71,7 @@ func (ls Labels) Filter(keep []string) LabelsIfc {
 }
 
 // convert Label set to a string in the form key1=v1,key2=v2.. + name + hash
```
gtopper, line 73: Need to update or delete comment.

pkg/utils/labels.go
```diff
@@ -195,6 +197,25 @@ func (ls Labels) Hash() uint64 {
 	return hash.Sum64()
 }
 
+// Hash returns a hash value for the label set.
```
gtopper, line 200: Need to update or delete comment.

gtopper (approved):
Only comments about comments.

PR #542, opened 2020-11-10
IG 17083 lru fixes

dinal:


pkg/appender/ingest.go
```diff
@@ -241,6 +241,9 @@ func (mc *MetricsCache) postMetricUpdates(metric *MetricState) {
 		if !sent {
 			if metric.store.samplesQueueLength() == 0 {
 				metric.setState(storeStateReady)
+				if metric.store.numNotProcessed == 0 && metric.store.pending.Len() == 0 {
```
talIguaz, line 244: `metric.store.pending.Len() == 0` is the same as `metric.store.samplesQueueLength() == 0`

pkg/appender/lru_cache.go
```diff
@@ -73,7 +64,7 @@ func (c *Cache) removeOldest() {
 		ele := c.free.Back()
 		if ele != nil {
 			c.free.Remove(ele)
-			kv := ele.Value.(*clist.Element).Value.(*entry)
+			kv := ele.Value.(*entry)
```
talIguaz, line 67: you can save the `key` itself, you don't use `kv`

talIguaz: Also, isn't `ele` can be nil after line 66? better get the key before the remove

pkg/appender/lru_cache.go
```diff
@@ -86,7 +77,8 @@ func (c *Cache) ResetMetric(key uint64) {
 	defer c.mtx.Unlock()
 	if ele, ok := c.cache[key]; ok {
 		c.used.Remove(ele)
-		c.free.PushFront(ele)
+		c.free.Remove(ele)
+		c.cache[key] = c.free.PushFront(&entry{key, ele.Value.(*entry).value})
```
talIguaz, line 81: Isn't `ele` nil after both Removes?

pkg/appender/lru_cache.go
```diff
@@ -35,21 +35,15 @@ func NewCache(max int) *Cache {
...
-		//check if element was already in list and if not push to front
-		c.used.MoveToFront(ee)
-		if c.used.Front() != ee {
-			c.used.PushFront(ee)
-		}
-		ee.Value.(*entry).value = value
-		return
+	if ele, ok := c.cache[key]; ok {
+		c.free.Remove(ele)
+		c.used.Remove(ele)
```
talIguaz, line 40: In my opinion better add a comment explaining why you delete from both lists, and that you expect one to "fail"

pkg/appender/lru_cache.go
dinal: ele.Value won't be nil, remove only deletes the list associated vars (head, prev, next)

talIguaz (approved):

